PR #46, opened 2013-10-20
Select parameter type, set application name, Czech locale

PetrDlouhy:
First patch introduces third optional dictionary parameter to the CONSTANCE_CONFIG tuple. Through this it allows user to set possible choices that can be selected in admin constance table. For such attributes the forms.Select widget is used.

Second patch allows to set name of the virtual application and model in admin.

Third patch is just Czech locale.


PetrDlouhy:
@bufke: Hi, sorry for the delay. I tried different approach, which allows to set custom fields with customary widgets in 8fdbf41519fd840d7ab0d58a8bce98b8e480bb85. It is also simpler and more understandable in config.

Can you check that, before I make PR? Please write that you received this, so I know that I should wait for your review.

PetrDlouhy:
@bufke: I made the PR (#108). Can you check that? Have you any idea, why is the Travis failing?

bufke:
Hello @PetrDlouhy sorry for the delay. I no longer have write access to this repo so I cannot help you. I do maintain the django-constance-updated pypi name space but just as a temporary measure of convenience for 1.8 support. 

I would suggest making translations and features as separate pull requests. Translations are an easy accept but adding new features is not.

I don't think @jezdez is interested in new features or making django-constance meet every need ever - which is totally fair. A friendly fork is not an bad thing - I would suggest doing so. I could help maintain but I don't have time to be leading such an effort or really doing active development here.

PetrDlouhy:
@bufke: The answer is under the PR #108

PR #47, opened 2013-11-14
ReadOnly Setting and Caching of backend results

trbs:
Added a READONLY settings for constance.

This is helpful in the case where for example a cluster of webservers are behind a PostgreSQL, Redis or other readonly backend slaves.

The setting avoids trying to save a setting if it does not exist in the (slave) backend.

----

Added CachedConfig and CachedAllConfig classes.

The use case for these are websites with a lot of requests per second. While profiling we found that when your RPS is getting high enough a decent amount of time is being spend doing Redis calls. Specially if there many different constance values to be checked for every request.

CachedConfig will cache an individual option for CACHE_TIMEOUT seconds. CachedAllConfig will use the backends mget() method to retrieve all the options every time it refreshes the cache. The latter may be particularity useful when most of the options are actually used and therefor mget is more efficient then multiple single calls to get().

Note that these changes are not Thread-Safe. But I figured Constance is not particularly thread-safe anyways.

---

The pull request is not clean since there a other changes to my fork, my apologies for that. I doubt a clean pull request will be possible from my repository. However the changeset for the READONLY settings should apply cleanly.

If there is interest enough in these patched I can try to port them directly so they will merge cleanly.


jezdez:
@trbs I'm closing this PR since it has diverged too much from the main repo but feel free to open new PRs for the read-only feature or whatever you consider worth porting upstream again.

trbs:
No problem. I'm considering moving back or rebasing on ```django-constance``` again. If I do that then I'll create PR's for any missing feature that might be interesting for the upstream.

PR #92, opened 2015-02-17
Helpful error message for South users

hso:
Add code to display a message with instructions on how to enable South migrations for those still using it.


hso:
That is, for users stuck with South < 1.0

jezdez:
Yeah, this is a no-go for now as it doesn't take South's version into account. Just update to South 1.x already.

PR #123, opened 2015-12-23
Replace removed gif icons with svg versions

federicobond:
The new admin style for Django 1.9 has replaced its gif icons for svg versions.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -127,6 +127,7 @@ class ConstanceAdmin(admin.ModelAdmin):
             'opts': Config._meta,
             'form': form,
             'media': self.media + form.media,
+            'icon_type': 'gif' if VERSION < (1, 9) else 'svg',
         }
         for name, (default, help_text) in settings.CONFIG.items():
             # First try to load the value from the actual backend
--- a/constance/templates/admin/constance/change_list.html
+++ b/constance/templates/admin/constance/change_list.html
@@ -74,9 +74,9 @@
                         </td>
                     <td>
                         {% if item.modified %}
-                            <img src="{% static 'admin/img/icon-yes.gif' %}" alt="{{ item.modified }}" />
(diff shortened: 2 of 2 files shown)


camilonova:
@federicobond Can you think of a way to support Django < 1.9 ?

federicobond:
It's easy to copy those files to the right locations if you are using Django < 1.9. Anything else is probably overkill.

camilonova:
If I merge this people having django < 1.9 will be affected, since django 1.8 is LTS I think makes sense to support out of the box the icons, without needing to do anything manually.

@jazzband/members Can anyone else help me reviewing a proper solution for this?

pjdelport:
> @jazzband/members Can anyone else help me reviewing a proper solution for this?

Would it make sense for `ConstanceAdmin` to determine the right icon paths based on the Django version, and pass them in as part of the template context?

albertyw:
@pjdelport That could work, but that would also mean having to check that path before every Django release, in case it gets changed again.  

I see a few other options:
 - Have some code, javascript or python, find the file and generate a path to use
 - Dive into django admin's code and find whatever function or variable that holds the path and call that.  
 - Make this repo contain a local copy of those two images
 - Just use text instead of those images.  They're files, but you could consider them as "private variables" that shouldn't be read from outside of django admin

I think I'd try (2) first, then go for your @pjdelport's suggestion or (3).

michaelkuty:
if you have some time to spare, please try to solve important things..

pjdelport:
> that would also mean having to check that path before every Django release, in case it gets changed again.

That should be acceptable: Django releases are infrequent, and Django admin redesigns are even less frequent.

federicobond:
If there is consensus on this, I can update the pull request with @pjdelport's proposal to determine the right icon paths based on the Django version.

pjdelport:
:+1:

federicobond:
Done.

camilonova:
@federicobond what an elegant solution. Thank you.

PR #130, opened 2016-02-22
Moved ConstanceForm to a property to make it exchangable by inheritance

mpauly:
I moved the form class for constance to a property of the Admin, similar to the form property in the default Model Admin, but gave it a different name in case the actual form will be used in some kind of detailed settings view in the future.
This allows to easily customize the Form used by Constance for rendering the settings, by simply subclassing ConstanceAdmin and ConstanceForm.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -112,6 +112,7 @@ class ConstanceForm(forms.Form):
 
 class ConstanceAdmin(admin.ModelAdmin):
     change_list_template = 'admin/constance/change_list.html'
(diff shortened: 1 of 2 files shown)


camilonova:
@mpauly Can you please document how to use it. I really like it.

mpauly:
Okay, I just added some documentation, I think from how ModelForms are normally handled in Django Admin it should be pretty obvious how to use this. Thanks for considering this pull request so quickly!

camilonova:
Thank you

PR #133, opened 2016-03-12
Fix formatting of CONSTANCE_ADDITIONAL_FIELDS doc

therefromhere:
Code examples aren't rendering currently on http://django-constance.readthedocs.org/en/latest/#custom-fields due to missing line at the start of code-blocks.


camilonova:
Thank you

PR #136, opened 2016-03-29
Typo on the variable name of verification hash version

emilioramirez:
The name of the variable has missing "CONSTANCE_" string at the begining


camilonova:
Thanks

PR #137, opened 2016-04-05
Use SplitDateTimeField instead of DateTimeField

jonzlin95:
https://docs.djangoproject.com/en/1.9/releases/1.9/#removed-features-1-9

The ability to use a SplitDateTimeWidget with DateTimeField is removed.

Update to use SplitDateTimeField instead. Cheers.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -109,7 +109,7 @@ class ConstanceForm(forms.Form):
     def clean_version(self):
         value = self.cleaned_data['version']
 
-        if settings.CONSTANCE_IGNORE_ADMIN_VERSION_CHECK:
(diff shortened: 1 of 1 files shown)


jonzlin95:
Also fix a setting name which is wrong in admin.py

fission6:
In case anyone finds this and wants a temporary fix for django 1.9

you can add the following in your settings

```
import datetime

CONSTANCE_ADDITIONAL_FIELDS = {
    datetime.datetime: ['django.forms.fields.SplitDateTimeField', {'widget': 'django.contrib.admin.widgets.AdminSplitDateTime'}],
}
```

PR #138, opened 2016-04-11
Undo last commit as it introduces an error

mpauly:
The last merged pull request #136  introduced an error, as in line 112 it used the prefixed version of `IGNORE_ADMIN_VERSION_CHECK` whereas it should used the one without prefix as defined in settings.py. Reverting the last commit fixes the error.


mpauly:
Also see the last pull request #137 , which also fixes the same error. You could alternatively merge that one.

PR #159, opened 2016-09-14
Moved the signal to its own file to avoid import errors. Fixes #158

### LINKED ISSUE #158: provide signal sending on "save"
### Describe the problem

**Feature request:**
Please provide signal sending on "save" (similar to post_save in django models) so user app could catch it and perform some actions with just updated settings

### Steps to reproduce

N/A

### System configuration

* Django version: 1.10
* Python version: 2.7

camilonova:
I tried to add tests but seems you can only tests signals using Mock.

Please let me know if anything else is needed.

### DIFF
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -86,7 +86,7 @@ class DatabaseBackend(Backend):
         if self._cache:
             self._cache.set(key, value)
 
-        signals.updated_signal.send(sender='constance', key=key, value=value)
+        signals.config_updated.send(sender='constance', key=key, value=value)
 
     def clear(self, sender, instance, created, **kwargs):
         if self._cache and not created:
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -4,8 +4,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db.models.signals import post_save
 
 from .. import Backend
-from ... import settings
-from ... import updated_signal
+from ... import settings, signals
 
 
 class DatabaseBackend(Backend):
@@ -87,7 +86,7 @@ class DatabaseBackend(Backend):
         if self._cache:
             self._cache.set(key, value)
 
-        updated_signal.send(sender='constance', key=key, value=value)
+        signals.updated_signal.send(sender='constance', key=key, value=value)
 
     def clear(self, sender, instance, created, **kwargs):
         if self._cache and not created:
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -4,7 +4,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db.models.signals import post_save
 
 from .. import Backend
-from ... import settings, signals
+from ... import settings, signals, config
 
 
 class DatabaseBackend(Backend):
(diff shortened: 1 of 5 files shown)


constance/backends/database/__init__.py
```diff
@@ -87,7 +86,7 @@ def set(self, key, value):
         if self._cache:
             self._cache.set(key, value)
 
-        updated_signal.send(sender='constance', key=key, value=value)
+        signals.config_updated.send(sender='constance', key=key, value=value)
```
jezdez: Passing in a `sender` parameter as a string doesn't make sense here, since it's used for [filtering down callbacks](https://docs.djangoproject.com/en/1.10/topics/signals/#connecting-to-specific-signals) in case a signal is used for different type of domain objects.

E.g. in the case of Django's model class signals like `post_save`, the signal is sent for no matter which model class. The difference between the call is the sender parameter.

Constance doesn't have multiple domain objects, so leaving the parameter to the default of `None` is better suited and even simplifies the developer facing API.

As a developer instead of using `updated_config.connect(sender='constance', config_changed_callback)` it's suffices to just use `updated_config.connect(config_changed_callback)`.

On the other hand, often signal callbacks are at risk of causing circular imports, and may not want to import the `constance.config` object in their module. In addition it makes sense for the developer who writes the signal callback to get something sent over that facilitated the config value change.

That's why I think it'd be best to pass in the `constance.config` object as the `sender` parameter, and document that that's what connected signal callbacks will receive as their first parameter. That way developers can created callback functions that are isolated and can act upon config changes by accessing the config object again as well.

IOW, this line should be `signals.config_updated.send(sender=config, key=key, value=value)`, where `config` is the lazy config objects `constance.config`.

constance/backends/redisd.py
```diff
@@ -50,4 +49,4 @@ def mget(self, keys):
 
     def set(self, key, value):
         self._rd.set(self.add_prefix(key), dumps(value))
-        updated_signal.send(sender='constance', key=key, value=value)
+        signals.config_updated.send(sender='constance', key=key, value=value)
```
jezdez: Same as above, this just can have `sender=config`.

docs/index.rst
```diff
@@ -72,15 +72,15 @@ Use this option in order to skip hash verification.
...
 
 .. code-block:: python
 
-    from constance import updated_signal
+    from constance.signals import config_updated
 
-    @receiver(updated_signal)
+    @receiver(config_updated)
     def constance_updated(sender, key, value, **kwargs):
         print(sender, key, value)
```
jezdez: Please add a sentence noting that sender here is the `config` object.

constance/backends/database/__init__.py
camilonova: How should I get the `config` object in this file?

I'm guessing I should do something like:
```
from constance import config
...
signals.config_updated.send(sender=config)
```

In case you want to see the other parameters should I rename them to `updated_key` and `new_value` ?
```
signals.config_updated.send(sender=config, updated_key=key, new_value=value)
```

jezdez: Yeah, the import looks right, and the new parameter names as well, good idea.

jezdez: Oh, I should we pass in the **old** value as well?

camilonova: I'll be hard to get the **old** value as it will add another hit to the backend. I believe for now only getting notified about the change is best.

jezdez: Ah, good point, it's cheap for the redis backend, but for the db backend you're right

docs/index.rst
```diff
@@ -81,11 +81,15 @@ You can use it as:
...
 
     @receiver(config_updated)
-    def constance_updated(sender, key, value, **kwargs):
-        print(sender, key, value)
+    def constance_updated(sender, updated_key, new_value, **kwargs):
+        print(sender, updated_key, new_value)
 
-In case you need it the sender is `constance`, and the key and value are
-the ones just changed.
+The sender is the `config` object, and the `updated_key` and `new_value`
```
jezdez: Just a nit, please use double backticks for inline code in rST

jezdez:
Just a minor nit for the docs, then this is good to go 👍 Thank you for working on this, @camilonova! 🌞

camilonova:
@jezdez Thank you for the patience 🌴

PR #160, opened 2016-09-14
Improve html layout when using fieldsets

camilonova:
I'm adding a fieldset structure like the one used in the inlines to improve readability when having multiple fields inside fieldsets.


jezdez:
LGTM!

PR #164, opened 2016-09-24
Backport command functonality from django-constance-cli

therefromhere:
Backporting command function from https://github.com/NZME/django-constance-cli in as requested by @jezdez in NZME/django-constance-cli#2

Basic usage:

```
./manage.py constance list
./manage.py constance get FOO
./manage.py constance set FOO bar
```

A couple of issues to discuss before merging:
- ~~Should it be possible to disable this command functionality?   I could do this by moving the command to a sub-app (eg constance.cli), or with a setting?~~
  *\* I'm not sure it's necessary to allow this to be disabled (eg, I don't think it's a security risk - if you've got manage.py access then you can do anything anyway).
- ~~The command line syntax could be expanded or altered, I've loosely modelled it on git config, any better suggestions?~~
  *\* I've updated as per discussion to positional arguments


tests/settings.py
```diff
@@ -64,7 +64,7 @@
     'LONG_VALUE': (long_value, 'some looong int'),
     'BOOL_VALUE': (True, 'true or false'),
     'STRING_VALUE': ('Hello world', 'greetings'),
-    'UNICODE_VALUE': (six.u('Rivière-Bonjour'), 'greetings'),
+    'UNICODE_VALUE': (u'Rivière-Bonjour', 'greetings'),
```
camilonova: It will work on python3 ?

therefromhere: It should be fine on 3.3+ , and I understand constance doesn't support 3.2 any more?  eg tox testing was removed https://github.com/jazzband/django-constance/commit/3fef92734927cefabccc9b7eab32207e8bbe306a

It seemed like with six.u() the test was sending wrongly encoded data before.

tests/test_cli.py
```diff
@@ -0,0 +1,97 @@
...
+
+        call_command('constance', set=('EMAIL_VALUE', 'blah@example.com'))
+
+        self.assertEqual(config.EMAIL_VALUE, "blah@example.com")
+
+    def test_get_invalid_name(self):
+        try:
+            call_command('constance', get=('NOT_A_REAL_CONFIG',))
+        except CommandError as e:
+            self.assertEqual(e.message, "NOT_A_REAL_CONFIG is not defined in settings.CONSTANCE_CONFIG")
```
jezdez: This would be a good use case for `assertRaisesMessage`

constance/forms.py
```diff
@@ -0,0 +1,133 @@
...
+            return value
+
+        if value != self.initial['version']:
+            raise forms.ValidationError(_('The settings have been modified '
+                                          'by someone else. Please reload the '
+                                          'form and resubmit your changes.'))
+        return value
+
+
+def get_constance_field(name):
```
jezdez: No need to add an extra function for this, please don't add unnecessary APIs to the app

constance/utils.py
```diff
@@ -1,6 +1,27 @@
...
+
+config = LazyConfig()
+
 
 def import_module_attr(path):
     package, module = path.rsplit('.', 1)
     return getattr(import_module(package), module)
+
+
+def get_constance_values():
```
jezdez: Please rename this to `get_values`

docs/index.rst
```diff
@@ -143,7 +143,7 @@ Note: Use later evaluated strings instead of direct classes for the field and wi
         }
 
 Ordered Fields in Django Admin
------------------------
+------------------------------
```
jezdez: Nice! 🍰

docs/index.rst
```diff
@@ -234,6 +234,31 @@ any other variable, e.g.:
...
+
+`--list` - output all values in a tab-separated format::
+
+    $ ./manage.py constance --list
+    THE_ANSWER 42
+    SITE_NAME  My Title
+
+`--get SETTING` - output a single values::
+
+    $ ./manage.py --get THE_ANSWER
```
jezdez: misses the name of the management command

docs/index.rst
```diff
@@ -234,6 +234,31 @@ any other variable, e.g.:
...
+    42
+
+`--set SETTING VALUE` - set a single value::
+
+    $ ./manage.py --set SITE_NAME "Another Title"
```
jezdez: misses the name of the management command

tests/test_cli.py
```diff
@@ -0,0 +1,97 @@
...
+    try:
+        yield new_target # run some code with the replaced stdout
+    finally:
+        sys.stdout = old_target # restore to the previous value
+
+
+class CliTestCase(TransactionTestCase):
+
+    def test_help(self):
+        out = StringIO()
```
jezdez: Just set `out` in a `setUp` method?

constance/forms.py
```diff
@@ -0,0 +1,133 @@
...
+    form = ConstanceForm(initial=get_constance_values())
+
+    return form.fields[name]
+
+
+def set_constance_value(raw_name, raw_value):
```
jezdez: Please prepend this with `_` to make sure people understand it's not an official API endpoint

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+    def add_arguments(self, parser):
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument("--list", action="store_true",
+                           help=_("List all Constance settings"))
+        group.add_argument("--get", dest='setting', nargs=1)
+        group.add_argument("--set", dest='setting', nargs=2, metavar=('SETTING', 'VALUE',))
+
+    def handle(self, *args, **options):
+
+        if VERSION < (1, 8):
```
jezdez: Django < 1.8 isn't supported anymore, you can remove this

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+            # since we assume use of argparse
+            raise CommandError(_("Constance CLI requires Django 1.8 or later"))
```
jezdez: No need to mark management command output as translatable, that's an anti-pattern since CLI output should be parsable. Django doesn't mark those either.

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if options.get('setting'):
+            raw_name = options.get('setting')[0]
+
+            if len(options.get('setting')) == 1:
+                # get
+                try:
+                    print(getattr(config, raw_name))
+                except AttributeError as e:
+                    raise CommandError(raw_name + _(" is not defined in settings.CONSTANCE_CONFIG"))
```
jezdez: No i18n here.

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+            else:
+                # set
+                raw_value = options.get('setting')[1]
+
+                try:
+                    set_constance_value(raw_name, raw_value)
+                except KeyError as e:
+                    raise CommandError(raw_name + _(" is not defined in settings.CONSTANCE_CONFIG"))
```
jezdez: No i18n here.

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+                except ValidationError as e:
+                    raise CommandError(",".join(e))
```
jezdez: Space after the comma, please

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+    def handle(self, *args, **options):
+
+        if VERSION < (1, 8):
+            # since we assume use of argparse
+            raise CommandError(_("Constance CLI requires Django 1.8 or later"))
+
+        if options.get('setting'):
+            raw_name = options.get('setting')[0]
+
+            if len(options.get('setting')) == 1:
```
jezdez: Maybe fetch the setting from the `options` here once instead of having to redo that below again?

constance/management/commands/constance.py
```diff
@@ -0,0 +1,57 @@
...
+                try:
+                    set_constance_value(raw_name, raw_value)
+                except KeyError as e:
+                    raise CommandError(raw_name + _(" is not defined in settings.CONSTANCE_CONFIG"))
+                except ValidationError as e:
+                    raise CommandError(",".join(e))
+
+        elif options.get('list'):
+            for k, v in get_constance_values().items():
+                print(k, v, sep="\t")
```
jezdez: Please use `self.stdout.write` instead

constance/utils.py
```diff
@@ -1,6 +1,27 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+
 from importlib import import_module
 
+from constance import LazyConfig, settings
+
+config = LazyConfig()
```
jezdez: I think instantiating another config instance isn't needed here anymore since the reason it was done originally was to prevent a circular import in the admin module. You could in theory use the `constance.config` lazy instance instead

docs/index.rst
```diff
@@ -234,6 +234,31 @@ any other variable, e.g.:
...
+
+Note that validation is done as per the admin form, an error will be raised if validation fails.
```
jezdez: Mind making that a rST note with `.. note::`?

tests/test_cli.py
```diff
@@ -0,0 +1,97 @@
...
+from textwrap import dedent
+
+from django.core.management import call_command, CommandError
+from django.test import TransactionTestCase
+from django.utils.six import StringIO
+
+from constance import config
+
+@contextmanager
+def redirect_stdout(new_target):
```
jezdez: Hm, what does Django do for this? Does it contain a helper we can reuse to test the out of the management command?

tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+            'CHOICE_VALUE': 'yes',
+            'TIME_VALUE': datetime.time(23, 59, 59),
+            'DATE_VALUE': datetime.date(2010, 12, 24),
+            'LINEBREAK_VALUE': 'Spam spam',
+            'DECIMAL_VALUE': Decimal('0.1'),
+            'STRING_VALUE': 'Hello world',
+            'UNICODE_VALUE': u'Rivière-Bonjour',
+            'DATETIME_VALUE': datetime.datetime(2010, 8, 23, 11, 29, 24),
+            'LONG_VALUE': 123456
+        })
```
jezdez: New line at the end of the file please

therefromhere:
Another issue I've found is that the `--set` argument seems to be clashing with Django's own `--setting` argument, if I can't resolve this I'll change the argument name (or make it a positional argument).

jezdez:
@therefromhere Yeah, that seems sensible, `./manage.py constance set FOO bar`

BTW I think we should warn people in the docs to use quotes for the value in case it contains spaces. Alternatively I could also see `./managepy constance set FOO=bar` to be the syntax, and then cut off at the first = and treat the rest of argv as the value?

constance/utils.py
therefromhere: I was having some problems with circular imports, I'll see if I can tidy this up

therefromhere:
A bit factor in the large number of changes is I split the form out out admin, but I can see if I can move it back to reduce the size of the delta.

tests/test_cli.py
therefromhere: I've removed

tests/test_cli.py
therefromhere: done

tests/test_cli.py
therefromhere: done

constance/forms.py
therefromhere: removed

constance/forms.py
therefromhere: done, moved to command

constance/utils.py
therefromhere: done

docs/index.rst
therefromhere: oops, fixed

constance/management/commands/constance.py
therefromhere: removed

constance/management/commands/constance.py
therefromhere: done

constance/management/commands/constance.py
therefromhere: refactored

therefromhere:
@jezdez can you take another look, I think this is pretty much ready to go now, the only issue I'm aware of being an extra LazyConfig().

I've reduced the size of the changes a lot by not splitting up admin.py.

camilonova:
@therefromhere Please check again, you have more changes from the last review to be done.

docs/index.rst
therefromhere: done

docs/index.rst
therefromhere: fixed

constance/management/commands/constance.py
therefromhere: fixed

constance/management/commands/constance.py
therefromhere: fixed

constance/management/commands/constance.py
therefromhere: fixed

constance/management/commands/constance.py
therefromhere: fixed

tests/test_utils.py
therefromhere: fixed

tests/settings.py
therefromhere: https://pythonhosted.org/six/#six.u

> On Python 2, u() doesn’t know what the encoding of the literal is. Each byte is converted directly to the unicode codepoint of the same value. Because of this, it’s only safe to use u() with strings of ASCII data.

So if we do need to support Python 3.2 in tests we can us six.u(), but with escape sequences instead of the accented characters.

therefromhere:
@camilonova , I think I've resolved all the issues?

This is the first large-ish PR I've done on github though - is there a better way to mark requested changes from the review as done other than replying to each comment individually with "fixed" etc?

camilonova:
@therefromhere Lets wait for @jezdez

therefromhere:
@jezdez , can you take another look?

camilonova:
@therefromhere John could you please squash all commits and create a new pull requests so it would be easier for @jezdez to take a look?

therefromhere:
@camilonova will do. It seems a bit more work is needed anyway since this branch's tests are failing after updating from master.

camilonova:
@therefromhere Thank you.

therefromhere:
@camilonova done as #180

PR #165, opened 2016-09-25
Updated example app to django 1.8, added migrations

therefromhere:
The example seemed to be out of date, I've regenerated the settings based on Django 1.8 defaults, and added migrations.


camilonova:
Thank you

PR #166, opened 2016-10-01
Fixed wrong sorting for config. Added sorting for config fieldset

Lumax-Rus:
Fixed wrong sorting for config. Added sorting for config fieldset


camilonova:
Thanks.

PR #168, opened 2016-10-05
Fixed #162. Made travis configuration more readable.

### LINKED ISSUE #162: Condition to release on travis is not working
### Describe the problem

By mistake I created a release without checking first if the tests passed. Still we have:

```
condition: "$TOXENV=py27-django-18"
```

on https://github.com/jazzband/django-constance/blob/master/.travis.yml#L45 and it should release a deploy only if the condition is true which was not in this case.

https://travis-ci.org/jazzband/django-constance/builds/160722075

Can anyone tell if this is not setup properly?

felixxm:


camilonova:
@felixxm Thank you for your interest in fixing the issue.

When you change the travis configuration it looks like Travis don't show the exact python version used to run the test. Take for example a current build vs the one you propose:

screen shot 2016-10-05 at 3 02 38 pm

As you can see on the left it shows the Python version used and with your changes it doesn't.

Also you say it fixes the #162 issue. Can you elaborate more why this solves this issue?

felixxm:
As you can see in travis [doc](https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A) build condition will be pasted into a bash expression:
```bash
if [[ <condition> ]]; then <deploy>; fi
```
In `bash` string comparison work only with whitespaces framing `=`. `"$TOXENV=py27-django-18"` (without whitespaces) will the translate to string ``"py27-django-18=py27-django-18"`` which is always `true`.

camilonova:
Great. Thanks.

How about the other changes from line 9 to 21 ?

felixxm:
Travis displays default python version, inside test it will use appropriate. You can check travis-ci issue about python3.5 (see [#4794](https://github.com/travis-ci/travis-ci/issues/4794)) which is not easily available in travis env. The easiest solution is to set it as a default python version, just like for example in django-debug-toolbar ([.travis.yml](https://github.com/jazzband/django-debug-toolbar/blob/master/.travis.yml)).

camilonova:
Thank you so much.

PR #171, opened 2016-10-11
Add django app config for "constance.backends.database"

lexqt:
It is not cool to have app "database" (default name is same as module name) in django registry.

### DIFF
--- a/constance/backends/database/apps.py
+++ b/constance/backends/database/apps.py
@@ -2,7 +2,7 @@ from django.apps import AppConfig
 from django.utils.translation import ugettext_lazy as _
 
 
-class ConstanceConfig(AppConfig):
+class ConstanceDatabaseConfig(AppConfig):
     name = 'constance.backends.database'
     label = 'constance_db'
     verbose_name = _('Constance Database Backend')
--- a/constance/backends/database/apps.py
+++ b/constance/backends/database/apps.py
@@ -1,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
+
+
+class ConstanceConfig(AppConfig):
+    name = 'constance.backends.database'
+    label = 'constance_db'
+    verbose_name = _('Constance Database Backend')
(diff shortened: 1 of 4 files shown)


docs/backends.rst
```diff
@@ -89,15 +89,15 @@ make sure the data model is correctly created::
 
 Please make sure to apply the database migrations::
 
-    python manage.py migrate database
+    python manage.py migrate constance_db
```
jezdez: Won't this break the migrations system since you're basically moving the app? If yes, then this is a no-go since it'd break every deployed version of constance hard and would require faking the constance migrations again. Not worth the hassle.

constance/backends/database/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
+
+
+class ConstanceConfig(AppConfig):
+    name = 'constance.backends.database'
+    label = 'constance_db'
```
jezdez: See my comments below, please don't rename the label of the app since it'll break the already applied migrations of deployed versions of the app

lexqt: It may be solved this way
https://github.com/omab/python-social-auth/commit/361f42235fe51f422af029f325d041407a16036c

social auth had the same problem with app name "default" (even worse than "database")

jezdez: Good idea, can you try this out first?

constance/backends/database/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
+
+
+class ConstanceConfig(AppConfig):
```
jezdez: Ah, one other thing, please rename the app config class to `ConstanceDatabaseConfig` to not confuse it with the other instances of "config" elsewhere in the app

constance/backends/database/apps.py
lexqt: Changed "app" to "database" in `django_migrations` table. `replaces` works.
```
$ ./manage.py migrate constance_db
Operations to perform:
  Apply all migrations: constance_db
Running migrations:
  No migrations to apply.
$ ./manage.py showmigrations constance_db
constance_db
 [X] 0001_initial (1 squashed migrations)
```

PR #172, opened 2016-10-27
Add app_config property to Config.Meta

miguelgr:
I found an scenario working with other django projects where the Meta for a model needs to have the `app_config` property as any other django.db.models.Model.Meta.


camilonova:
@jezdez what do you think?

camilonova:
I'll take a chance on this. Seems it will not add any issue.

@miguelgr Gracias 🍺

PR #177, opened 2016-11-21
Preserve line breaks in default value

rvernica:
In the default value of a field is of type string and has line breaks (using `\n`), they are not preserved when the *Constance config* page is rendered in the Django Admin. This patch adds the [`linebreaks`](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#linebreaks) filter to the template so that the newlines ('\n') are converted to line break ('\<br\>').


camilonova:
Thank you

PR #179, opened 2016-11-27
Additional fields fix

therefromhere:
- Don't ignore additional fields when creating form.  Fixes regression from 2f88a1bff2d3741cd18e1b47fe8c8c1d882795f4
- Added test of form field types
- Added CONSTANCE_ADDITIONAL_FIELDS to example app
- Updated ChoiceField example, the key & value seemed to be the wrong way around?


camilonova:
Thank you

PR #191, opened 2017-01-10
Fix *Reset to default* to work with boolean/checkboxes

### LINKED ISSUE #189: "Reset to default" not working on boolean widget
### Describe the problem

The reset to default link isn't working for checkboxes

Because the generated JS is:

`document.getElementById('id_MY_BOOLEAN').value = 'True'; return false;`

There's 2 issues here:  

* stringified python `'True'` instead of JS bool `true`
* Checkboxes don't have a `.value`, they used `.checked`

This would work:

`document.getElementById('id_MY_BOOLEAN').checked = true; return false;`

### Steps to reproduce

Add a boolean Constance value.

### System configuration

* Django version: 1.8.17
* Python version: 2.7.12

rvernica:
* Add field name to each config value
* Check field name and use *checked* for checkbox and *value* otherwise

Fix #189

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -177,6 +177,7 @@ class ConstanceAdmin(admin.ModelAdmin):
             'value': localize(value),
             'modified': localize(value) != localize(default),
             'form_field': form[name],
+            'form_field_name': form[name].field.widget.__class__.__name__,
         }
 
         return config_value
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -173,11 +173,13 @@ class ConstanceAdmin(admin.ModelAdmin):
         config_value = {
             'name': name,
             'default': localize(default),
+            'raw_default': default,
             'help_text': _(help_text),
             'value': localize(value),
             'modified': localize(value) != localize(default),
             'form_field': form[name],
-            'form_field_name': form[name].field.widget.__class__.__name__,
+            'is_checkbox': isinstance(
+                form[name].field.widget, forms.CheckboxInput),
         }
 
         return config_value
--- a/constance/templates/admin/constance/includes/results_list.html
+++ b/constance/templates/admin/constance/includes/results_list.html
@@ -21,8 +21,8 @@
             {{ item.form_field }}
             <br>
             <a href="#" onClick="
-              document.getElementById('{{ item.form_field.auto_id }}').{% if item.form_field_name == 'CheckboxInput' %}checked =
-                  {% if item.default == 'True' %} true {% else %} false {% endif %}
+              document.getElementById('{{ item.form_field.auto_id }}').{% if item.is_checkbox %}checked =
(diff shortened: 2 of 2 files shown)


constance/admin.py
```diff
@@ -177,6 +177,7 @@ def get_config_value(self, name, options, form, initial):
             'value': localize(value),
             'modified': localize(value) != localize(default),
             'form_field': form[name],
+            'form_field_name': form[name].field.widget.__class__.__name__,
```
jezdez: FWIW, that's not really the "form field name" but the name of the form field's widget class.
In Django this is solved [differently](https://github.com/django/django/blob/6252fd6314f66d2d303cc47c791ffefd27169b42/django/contrib/admin/helpers.py#L138), so let's follow the example:

```python
'is_checkbox': isinstance(form[name].field.widget, forms.CheckboxInput)
```

Let's pass the value to the template instead of implementing it on template level. That also makes sure we don't hard code the class name, while 3rd party theming apps may monkeypaatch the `forms.CheckboxInput` with an own class with a different name.

constance/templates/admin/constance/includes/results_list.html
```diff
@@ -20,7 +20,10 @@
             {{ item.form_field.errors }}
             {{ item.form_field }}
             <br>
-            <a href="#" onClick="document.getElementById('{{ item.form_field.auto_id }}').value = '{{ item.default|escapejs }}'; return false;">Reset to default</a>
+            <a href="#" onClick="
+              document.getElementById('{{ item.form_field.auto_id }}').{% if item.form_field_name == 'CheckboxInput' %}checked =
+                  {% if item.default == 'True' %} true {% else %} false {% endif %}
```
jezdez: I'm confused why we need to do a string comparison here, if the default is a boolean, shouldn't we check here if it's truthy?

rvernica: I'm not very comfortable with the string comparison, but `item.default` is not a Boolean, but a String since it is passed through `localize` in [line 175](https://github.com/jazzband/django-constance/blob/81b723c91b7b1e708368c6d7cd04dc54e0abb878/constance/admin.py#L175):
```python
'default': localize(default),
```

therefromhere: should we have a raw_default (or similar) attribute to item to use here?

jezdez: @rvernica Ah, I missed that, good point. Then let's follow @therefromhere's idea and pass the raw default value as well to make the check. I can't come up with a better name, `raw_default` sounds good though.

rvernica:
Thanks for the comments. I updated the PR accordingly.

jezdez:
@rvernica Thanks!

Issue #185, opened 2016-12-23
Reset to default not working on datetime widget


camilonova:
### Describe the problem

The reset to default link is not working when having a datetime widget.

![screen shot 2016-12-23 at 5 59 26 pm](https://cloud.githubusercontent.com/assets/202559/21463617/c535e26e-c939-11e6-9122-ac50c77518c4.png)

I guess you have to split the value and convert it, since the value shown is different than the one used at the input.

### Steps to reproduce

Add a datetime as a constance value

### System configuration

* Django version: 1.10
* Python version: 3.5

camilonova:
@rvernica Can you please take a look?

rvernica:
OK, will do.

camilonova:
Maybe the best think would be to disable the reset feature if we have more than one input widget.

rvernica:
For `DateTime` type, instead of one field (`ID`) that we have to populate, there are two fields, one with `ID_0` and one with `ID_1`. Populating them should not be hard... The problem is getting the default. The default that is available in HTML is already localized and may be hard to split.

The current code does not even work for `Date` or `Time` types because the default available in HTML is localized:
![date-constance](https://cloud.githubusercontent.com/assets/241185/21486708/cea70400-cb6d-11e6-8b5b-c5f7ea9ac354.png)

A solution that would help with all three types would be to provide a raw default to the HTML view. For `DateTime` this default could be a tuple of two values, one for each part. Fixing the HTML and JavaScirpt should be easy once appropriate values as available.

jezdez:
Hm, I wonder if we should revert the reset to default feature, it kind of has evolved in all kinds of UI problems, which hurts constance reputation as a simple and stable tool. Anyone care to weigh in on this?

rvernica:
I might be biased but I think the feature is useful. I see its usefulness in one of the projects where I use constance. I think the issues can be addressed.

jezdez:
@rvernica Oh it's definitely useful, don't get me wrong, just I wondered if we need more time fleshing out the feature before re-adding it again? I'm also happy if you think we can figure out the various edge cases.

rvernica:
See PR #192 for a fix to this issue.

SebCorbin:
This seems to be solved, can't reproduce with last version

PR #192, opened 2017-01-19
Fix Reset to default for DateTime, #185

rvernica:
* I added a flag in the context for `DateTime` type
* In the template I use `get_format('DATE_INPUT_FORMATS')[0]` and `get_format('TIME_INPUT_FORMATS')[0]` from Django Admin to format the date and time. I similar pattern is used in for *Now* and *Today* functionality available in the widget.

### DIFF
--- a/constance/templates/admin/constance/includes/results_list.html
+++ b/constance/templates/admin/constance/includes/results_list.html
@@ -22,17 +22,23 @@
             <br>
             <a href="#" onClick="
               {% if item.is_datetime %}
-                d = new Date(
+                var defaultDate = new Date(
                   {{ item.raw_default|date:"Y" }},
                   {{ item.raw_default|date:"n" }} - 1,
                   {{ item.raw_default|date:"j" }},
                   {{ item.raw_default|time:"H" }},
                   {{ item.raw_default|time:"i" }},
                   {{ item.raw_default|time:"s" }});
+                var body = document.getElementsByTagName('body')[0];
+                var serverOffset = body.getAttribute('data-admin-utc-offset');
+                if (serverOffset) {
+                  var defaultOffset = defaultDate.getTimezoneOffset() * -60;
+                  defaultDate.setTime(defaultDate.getTime() + 1000 * (serverOffset - defaultOffset));
+                }                              
                 document.getElementById('{{ item.form_field.auto_id }}_0').value =
-                  d.strftime(get_format('DATE_INPUT_FORMATS')[0]);
+                  defaultDate.strftime(get_format('DATE_INPUT_FORMATS')[0]);
                 document.getElementById('{{ item.form_field.auto_id }}_1').value =
-                  d.strftime(get_format('TIME_INPUT_FORMATS')[0]);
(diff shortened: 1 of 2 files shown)


rvernica:
@jezdez are there any issues with this PR?

constance/templates/admin/constance/includes/results_list.html
```diff
@@ -21,9 +21,23 @@
...
-                  {% if item.raw_default %} true {% else %} false {% endif %}
-                {% else %}value = '{{ item.default|escapejs }}'{% endif %}; return false;">Reset to default</a>
+              {% if item.is_datetime %}
+                d = new Date(
+                  {{ item.raw_default|date:"Y" }},
+                  {{ item.raw_default|date:"n" }} - 1,
+                  {{ item.raw_default|date:"j" }},
+                  {{ item.raw_default|time:"H" }},
+                  {{ item.raw_default|time:"i" }},
+                  {{ item.raw_default|time:"s" }});
```
jezdez: This is ignoring a possible timezone attached to the datetime object. E.g. if the configuration value has been added with one in the settings, we need to make sure to respect that when instantiating the JavaScript Date object. That's especially important since we use `get_format` a bit below which can -- if `USE_L10N` is enabled -- render a timezone aware datetime in the current active timezone -- which depending on the project setup can and in my experience often is different from the server time.

Django solves this for the admin datetime widget by passing in the [UTC offset from the server side](https://github.com/django/django/blob/3c447b108ac70757001171f7a4791f493880bf5b/django/contrib/admin/templates/admin/base.html#L14-L15) and then uses it when it [constructs a new Date object](https://github.com/django/django/blob/3c447b108ac70757001171f7a4791f493880bf5b/django/contrib/admin/static/admin/js/admin/DateTimeShortcuts.js#L43-L53). I think we need to follow the same approach here to prevent accidentally setting the wrong value.

rvernica:
@jazdez would you mind having a look at this? I updated the code as you suggested but I am not sure it is correct to do such a transformation.

camilonova:
@jezdez Please take a look

camilonova:
@rvernica could you please update your files? I would like to see this merged.

rvernica:
The files were updated back then. There was no more feedback after that.

camilonova:
I'm seeing a conflict with one of the files. All the other feedback was done as I can tell.

Fixing the conflict will allow us to merge it, right?

rvernica:
I see. I wish this was merged back then... 

Resolving the conflict is not trivial. The way `Reset to default` is implemented has been changed in c8568b3f6b8098c70edacb1a91d41579686a46fb Maybe @whs can help.

In this patch we use two ID variables, `'{{ item.form_field.auto_id }}_0'` and `'{{ item.form_field.auto_id }}_1'`. The way `Reset to default` is implemented right now, it only accepts one.

whs:
I'll take a look later today after work.

camilonova:
Thank you. I hope we can sort this out today.

whs:
Continued in #243

PR #194, opened 2017-01-29
Fixed Django 2.0 tests due to `django.conf.urls.include` behavior change. Removed pypy for Django 2.0 tests.

felixxm:
Fixed Django 2.0 tests due to `django.conf.urls.include` behavior change i.e.
```
django.core.exceptions.ImproperlyConfigured: Passing a 3-tuple to django.conf.urls.include()
is not supported. Pass a 2-tuple containing the list of patterns and app_name, and provide
the namespace argument to include() instead.
```
Removed pypy for Django 2.0 tests. We cannot use `pypy3` instead (see https://github.com/travis-ci/travis-ci/issues/6277).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,12 @@ env:
   - TOXENV=pypy-django-19
   - TOXENV=pypy-django-110
   - TOXENV=pypy-django-111
-  - TOXENV=pypy-django-master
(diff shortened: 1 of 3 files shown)


camilonova:
Thank you

PR #195, opened 2017-01-31
Added the old value to the config_updated signal

camilonova:
One of the main uses of the config_updated signal is to act accordingly to field changes. Signal is already providing the key and the new value, but it misses the old value so anyone can compare if it was changed.

I'm proposing here to send into the signal the old value as well as the new value, so it allows to compare if the value was changed and act accordingly.

@jezdez Could you please take a look and let me know what you think?

### DIFF
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -77,6 +77,7 @@ class DatabaseBackend(Backend):
         return value
 
     def set(self, key, value):
+        old_value = self.get(key)
         constance, created = self._model._default_manager.get_or_create(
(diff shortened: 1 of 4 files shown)


camilonova:
I'm going to merge this one and see if anyone have something to say.

jezdez:
Ah, right, I glimpsed at it on the phone a few days ago and found it not problematic, although it is a backward incompatible change given the renaming of the signal handler arguments. If you release it, please bump the major version.

camilonova:
@jezdez yes sr.

PR #197, opened 2017-02-05
Fix create_perm in apps.py to use database alias given by the post_migrate signal

rlaszlo:
While running migrations with multiple databases I have noticed that the create_perm signal handler doesn't use the given "using" argument, which means that it always creates the ContentType and Permission in the default database.

The fix is to use the "using" argument on the objects manager with the using method. 

I have also added tests for this, which needed an extra "secondary" database in the settings.py.
This means that every db setup is done twice, which is unnecessary for every other tests, but I don't think that is possible to add a new database to the settings for only one test case.


camilonova:
Thank you

PR #198, opened 2017-02-14
Add a `get_changelist_form` hook in `ConstanceAdmin`

Kerl13:
This method is used in the `changelist_view` to get the changelist form.
    
You may want to override this method to serve different forms depending
on the user that make the request. For example:
    
    class MyConstanceAdmin(ConstanceAdmin):
        def get_changelist_form(self, request):
            if request.user.is_superuser:
                return SuperuserForm:
            else:
                return super(MyConstanceAdmin, self).get_changelist_form(request)

Django's `ModelAdmin` has similar hooks which are pretty useful when you need to allow different groups of users with different privileges to access the admin site.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -184,14 +184,23 @@ class ConstanceAdmin(admin.ModelAdmin):
 
         return config_value
 
+    def get_changelist_form(self, request):
+        """
+        Returns a Form class for use in the changelist_view.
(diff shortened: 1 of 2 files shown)


camilonova:
Thank you

Kerl13:
Thank your for your quick answer :)

PR #208, opened 2017-05-16
Add python 3.6 to build matrix

browniebroke:
I only have 2.7 and 3.6 on my machine, was helpful to get this in the tox.ini file. 
We need to set the python version explicitly on Travis, though, otherwise, it raises an interpreter not found error.


camilonova:
Thank you

PR #209, opened 2017-05-18
fix bug when site has multiple databases. 

sholleman:
ContentType now relys on the site's DB router instead of explicity setting a database via using().


sholleman:
Disregard. This is breaking tests so I have another idea for how to fix this.

PR #212, opened 2017-05-25
Fixed #211 -- Added datetime.timedelta support.

### LINKED ISSUE #211: Cannot store datetime.timedelta objects in config
### Cannot store `datetime.timedelta` objects in config
Although `python manage.py constance list` shows the value correctly, navigating to `constance` in Django admin throws **Server Error (500)**:
```
django.core.exceptions.ImproperlyConfigured: Constance doesn't support config values of the type <class 'datetime.timedelta'>. Please fix the value of 'VARIABLE'.
```

## Steps to reproduce

1. In config, define variable with type timedelta:
```
# (settings.py)
from datetime import timedelta
CONSTANCE_CONFIG = {
    'VARIABLE': (
        timedelta(days=5), 'Sample timedelta variable', timedelta
    ),
}
```
2. Navigate to `http://localhost:8000/admin/constance/config/`

### System configuration
* Django version: 1.10.4
* Python version: 3.6.1

felixxm:


camilonova:
Thank you

PR #215, opened 2017-06-02
Add testing for all supported Python and Django versions

jdufresne:
Document all supported versions in PyPI using trove classifiers.
Alphabetize classifiers.

Add all supported versions to tox.ini for easy testing. Tidy up tox.ini
by removing defaults for basepython.

Add all supported versions to the Travis CI configuration for CI
testing.

In the Travis and tox configuration use a more explicit test matrix as
the different versions of Django do not have complete overlap of Python
support.

Update Travis configuration to use built in pip caching support.

https://docs.travis-ci.com/user/caching/#pip-cache


.travis.yml
```diff
@@ -1,33 +1,63 @@
...
-  - TOXENV=py35-django-master
-  - TOXENV=pypy-django-18
-  - TOXENV=pypy-django-19
-  - TOXENV=pypy-django-110
-  - TOXENV=pypy-django-111
+cache: pip
+
 matrix:
   include:
+    - python: 2.7
```
jezdez: We should switch to [tox-travis](http://tox-travis.readthedocs.io/en/stable/) instead.

jdufresne:
Thanks for the tip. Updated to use Tox Travis. I wasn't aware of this before, neat package.

jezdez:
@jdufresne Ah, no problem. Yeah, it's quite neat. Since the Travis matrix is changes to be by language, could you update the [condition for the PyPI deploy](https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A) to `python: 3.6` instead of the custom condition `condition: "$TOXENV = py27-django-18"`?

jdufresne:
Done.

I don't normally use the `deploy:` feature, so please check that it looks correct.

jezdez:
@jdufresne Looks correct!

jezdez:
Thank you for the contribution!

PR #220, opened 2017-06-02
Fix for #219

### LINKED ISSUE #219: Wrong admin site used to generate links
With the newest version of constance it seems that in admin.py:215 the wrong admin site is used. Instead of `django.contrib.admin.site` one should probably use `self.admin_site`. Otherwise I experience problems of the following kind when using a custom admin site in combination with a fairly involved configuration of django admin: `NoReverseMatch: Reverse for 'app_list' with keyword arguments '{'app_label': 'someappnamethathasnoadmin'}' not found`

Django-constance: master
Django: 1.11.2

mpauly:
Use admin_site property of ModelAdmin, to make sure that the correct list of apps is used.
This should fix #219

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -212,7 +212,7 @@ class ConstanceAdmin(admin.ModelAdmin):
                 )
                 return HttpResponseRedirect('.')
         context = dict(
-            admin.site.each_context(request),
+            self.admin_site.each_context(request),
             config_values=[],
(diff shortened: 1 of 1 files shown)


mpauly:
@camilonova Any change this is merged? This is stopping me from using django-constance@master and should not break any existing installation as fas as I can tell.

camilonova:
Thank you.

PR #222, opened 2017-06-13
Fixed #201

### LINKED ISSUE #201: Forgetting variables in CONSTANCE_CONFIG_FIELDSETS makes the admin form unusable
### Describe the problem

If one forgets a variable in `CONSTANCE_CONFIG_FIELDSETS`, the constance admin interface fails to update the other fields with new values. After clicking on "Save", the values displayed just afterwards in the form are still the new ones, but they have not been saved on the database (my settings, but I guess it's the same with Redis), and the flag "modified" is not updated. 

The reason is [that the form is not valid](https://github.com/jazzband/django-constance/blob/484ab4113293349f880c70b964282a2dd6b3eae2/constance/admin.py#L204) because there is then a required filed is missing.

### Steps to reproduce

1. Have two constance variable, say `FOO`  and  `BAR`: 
```
CONSTANCE_CONFIG = {
    'FOO': (42, 'Answer to the Ultimate Question'),
    'BAR': (47, 'Answer to the Ultimate Question, taking into account Inflation'),
}
```` 
2. Forget `BAR` in `CONSTANCE_CONFIG_FIELDSETS`
```
CONSTANCE_CONFIG_FIELDSETS = {
    'Answers': ('FOO'),
}
```
3. Go to the constance admin page, and change the value of `FOO`. The new value is not saved, because the form is not valid because the field `BAR` is missing.

### Suggestion

Raise an error? Let me a couple of weeks and I'll have the time to do a PR

### System configuration

* Django version:  1.10.2
* Python version: 2.7.9

varunbargali:
…IELDSETS


varunbargali:
Added validation check to fix [#201](https://github.com/jazzband/django-constance/issues/201)

hurta2yaisel:
What about this?? What is the version of django-constance with these fixes?? I don't see any new release since February 17....

camilonova:
Thank you.

PR #224, opened 2017-07-03
Add French translations

browniebroke:
Not sure if that should be populated in Transiflex first? I couldn't find the French translations there, so I've used the below steps:

- `django-admin.py makemessages --locale fr` to generate the `po` file
- Edited the translations in [Poedit](https://poedit.net/)
- `django-admin.py compilemessages --locale fr` to generate the `mo` file


camilonova:
Thank you.

PR #225, opened 2017-07-05
Support config DB only

arybin93:
Parameter CONSTANCE_CONFIG_ONLY_DB was added.
This parameter is holding your variables in database via Constance DB, without display this values in admin panel.

 for using Constance DB without adding this parameters in user interface.

EXAMPLE FOR USING:

In settings:
CONSTANCE_CONFIG_ONLY_DB = {
    'YOUR_SETTINGS_TYPE': ('0', u'Description of settings', str),
    'YOUR_SETTINGS_TEXT': ('your text', u'Description of setting', str)
}

In your code:
get value: 
config.YOUR_SETTINGS_TYPE

set value: 
config.YOUR_SETTINGS_TYPE = '2'

### DIFF
--- a/constance/base.py
+++ b/constance/base.py
@@ -10,10 +10,15 @@ class Config(object):
             utils.import_module_attr(settings.BACKEND)())
 
     def __getattr__(self, key):
+        if key in settings.CONFIG_ONLY:
+            current_config = settings.CONFIG_ONLY
+        else:
+            current_config = settings.CONFIG
+
         try:
-            if not len(settings.CONFIG[key]) in (2, 3):
+            if not len(current_config[key]) in (2, 3):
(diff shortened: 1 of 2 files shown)


aleszoulek:
What is the use case for that? 

The reaison d'etre for Constance project was to be able to manipulate settings-like values over admin interface.

arybin93:
Yes, I got it and I like this idea in general. But for our project, we have many different settings-like values and we use Constance admin interface form like general settings and create new custom forms for another settings. This changes allow us used one model for all settings, but used different forms in admin interface.

Thank you for feedback.

camilonova:
I guess this is out of the scope of this library.

@arybin93 Thank you for your time.

I'll close this PR.

Issue #229, opened 2017-08-01
Unable to run migrations from scratch


DusanMadar:
### The problem
I have a project with 50+ migrations in it and recently decided I want to use Constance for a couple of settings.

Everything was OK while developing/adding dynamic settings. I guess it's due to the fact that I setup the Constance config first, did the migration and only then start to use dynamic settings values.

Although when I want to do a test deploy  (i.e. when starting up a new Vagrant VM) and run all the migrations from scratch it fails. I am unable to run `python manage.py migrate database`, `python manage.py migrate` or even `python manage.py showmigrations`.

```
  File "/vagrant/my_app/views.py", line 138, in MyViewSet
    my_variable = constance_config.MY_VARIABLE
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/utils/functional.py", line 239, in inner
    return func(self._wrapped, *args)
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/constance/base.py", line 19, in __getattr__
    result = self._backend.get(key)
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/constance/backends/database/__init__.py", line 74, in get
    value = self._model._default_manager.get(key=key).value
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py", line 374, in get
    num = len(clone)
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py", line 232, in __len__
    self._fetch_all()
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py", line 1118, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py", line 53, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch)
  File "/home/ubuntu/.virtualenvs/venv/lib/python3.5/site-packages/django/db/models/sql/compiler.py", line 894, in execute_sql
    raise original_exception
[...]

PR #230, opened 2017-08-07
database: Ignore operation errors (#229)

whs:
Ignore OperationErrors during mget/get to allow migration/test to run.

I intentionally left set unmodified, as it isn't needed during both operation. It should be a reminder to users that they need to run migration.


camilonova:
Thanks

aarcro:
Is this not released? I have v2.0.0 installed and have the trouble with ProgrammingError when running migrations on a fresh database.

camilonova:
Is not released yet. Please try constance on master and let me know if you think is ready for a release.

PR #231, opened 2017-08-09
Moved inline css/javascripts out to their own files

whs:
As constance use inline JavaScript to reset field to default, it will be unusable when the site have CSP setup. I've moved inline JS/CSS out to their own files.


camilonova:
Thank you

PR #243, opened 2017-11-07
Fix Reset to default for DateTime, #185

whs:
#192 updated to remove inline JavaScript


constance/templates/admin/constance/includes/results_list.html
```diff
@@ -20,9 +20,18 @@
...
+                data-field-id="{{ item.form_field.auto_id }}"
+                data-field-type="{% spaceless %}
+                {% if item.is_checkbox %}checkbox
+                {% elif item.is_datetime %}datetime
+                {% endif %}
+                {% endspaceless %}"
+                data-default="{% spaceless %}
                 {% if item.is_checkbox %}{% if item.raw_default %} true {% else %} false {% endif %}
-                {% else %}{{ item.default }}{% endif %}
+                {% elif item.is_datetime %}{{ item.raw_default|date:"U" }}
```
whs: I updated this to use POSIX time instead. I had a brief testing and that should be working, but my knowledge in timezone support is limited.

camilonova:
Thank you

PR #249, opened 2017-12-08
Support dynamic default values

simonschmidt:
When changing default values in settings.py it is sometimes desirable
for those changes to take effect for any unset config values.

As it currently works the default value is persisted to storage
when accessed.

This commit introduces a new setting:

CONSTANCE_DYNAMIC_DEFAULTS=False

That when set to True simply avoids saving the default value
to the storage backend.


simonschmidt:
One major issue I'd like some input on, is it possible for this type of thing to play well with the admin interface?

Since as soon as anything is saved there nothing will technically need the default value anymore.

camilonova:
I'm not sure if this should be in here, but I appreciate @jezdez thoughts on this.

jezdez:
Nope, it shouldn't. Django settings are by design changed not thread-safe and are cached during startup. Since Constance uses the Django settings system as the storage for the defaults we can't add a feature like this. If you need this in your application I suggest to store the values in a thread local in your app and make sure your code is able to read from there.

PR #256, opened 2018-02-25
Fix ConstanceForm validation

blablacio:
Account for cases where CONSTANCE_CONFIG_FIELDSETS is not set at all in ConstanceForm validation.


camilonova:
Thank you.

PR #257, opened 2018-02-27
add config label

legshort:
This modification is not only for a prevention of below error but also providing `model.label`.
I think this is not a mandatory or core PR but I think the model should have the label for later usage or compatible with others.

```python
File "/Users/legshort/.pyenv/versions/project/lib/python3.6/site-packages/admin_view_permission/utils.py", line 26, in get_model_name
    return model._meta.label
AttributeError: 'Meta' object has no attribute 'label'
```

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -292,6 +292,7 @@ class ConstanceAdmin(admin.ModelAdmin):
 
 class Config(object):
     class Meta(object):
+        label = 'constance.Config'
         app_label = 'constance'
         object_name = 'Config'
         model_name = module_name = 'config'


jezdez:
Could you make this more in line what [Django does](https://github.com/django/django/blob/6e52e2554dcc902b446f9d371ed1e6a07f36d6ec/django/db/models/options.py#L133-L139)? That'd help make sure any people monkey-patching this code path to prevent breakage.

```python
    @property
    def label(self):
        return '%s.%s' % (self.app_label, self.object_name)

    @property
    def label_lower(self):
        return '%s.%s' % (self.app_label, self.model_name)
```

Sinkler:
Hello! Recently I ran into the same problem and made a fork to fix this, please check #260

camilonova:
Closed in favor of #260

legshort:
I'm looking forward to releasing new version!

Issue #210, opened 2017-05-22
Migration error for multiple databases: django.db.utils.ProgrammingError: (1146, "Table 'db_2.django_content_type' doesn't exist")


sholleman:
### Describe the problem

Hi. We have a django site that uses 4 databases. I'm trying to install Constance on our Django site and am getting an error regarding the django_content_type table in one of the non-default databases. 

We've written a DB router to route selected apps to specific databases. Non-specified apps route to default. Because Django, we have a script to run the migrations for all the databases. 

```
python manage.py migrate --noinput 
python manage.py migrate --noinput --database=db_2 
python manage.py migrate --noinput --database=db_3 
python manage.py migrate --noinput --database=db_4 
```

Running the script gives the following error when it runs the db_2 command:

```  
Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying database.0001_initial... OK
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/sholleman/src/.virtualenvs/sma/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line

[...]

Migration error: django.db.utils.ProgrammingError: (1146, "Table 'db_2_staging.django_content_type' doesn't exist")
```

We considered creating a  `django_content_type` tables each database, but that feels wrong. Removing the two instances of `.using(using)` from create_perm() in apps.py stops the error, but breaks one of your tests, `test_post_migrate_signal_creates_content_type_and_permission_in_secondary_database`. I think the error is due to a disconnect between the migration and the post-migrate signal. The migration looks to our custom db router and doesn't do anything to db_2. But the create_perm() in apps.py is ignoring our db router. Do you have any suggestions for fixing this? 

### Steps to reproduce

- Have a multi-database site with a special DB mapping router to route model calls to the correct database. Only the default database has the table, `django_content_type`.
- Install Constance with database backend. 
- Run migrations for non-default database. e.g., `python manage.py migrate --noinput --database=ib_files`
- Error

### System configuration

* Django version: 1.8.15
* Python version: 2.7.12

jezdez:
First, I'm sorry you have 4 databases. I hope you understand that a out-of-the-box tool like constance doesn't often target this type of system when it's developed.

[...]

PR #258, opened 2018-03-02
Workaround to make `django-constance` work in multi-DB project

ownaginatious:
This is an attempt to fix the issue described in #210. What this PR does is add a `CONSTANCE_DBS` setting to the Django settings file. This is an interable that says which databases should be affected by the `django-constance` post-migration signals.

Not sure if this is the best way to fix this, but it seems to work for me ¯\\\_(ツ)_/¯

Totally open to better name suggestions than `CONSTANCE_DBS` :+1:


camilonova:
Thank you

PR #259, opened 2018-03-05
Add zh_Hans support for Django 1.9+

0bit:
In Django 1.9:

> The zh-cn and zh-tw language codes will be removed and have been replaced by the zh-hans and zh-hant language code respectively.

so copy `zh_CN` dir to `zh_Hans` dir.


0bit:
sorry, I think there has some problem to solve.

PR #264, opened 2018-03-13
Fix `CONSTANCE_CONFIG_FIELDSETS` mismatch issue

fengsi:
`CONSTANCE_CONFIG_FIELDSETS` should be optional (change in `2.1.0` broke it), otherwise nothing could be updated if `CONSTANCE_CONFIG_FIELDSETS` is not defined at all, which should be the most common use case.


camilonova:
Thank you

PR #276, opened 2018-04-19
update Chinese translation

yinkanghong:
It's not a simple copy of `zh_CN`,i remake the `zh_Hans/LC_MESSAGES/django.po` file and translated the `zh_Hans/LC_MESSAGES/django.po` file in more appropriate words.


camilonova:
Thank you

yinkanghong:
It's my pleasure

PR #278, opened 2018-04-20
fix bug of can't change permission chang_config's name

yinkanghong:
If i change the permission `change_config`'s name to other name and `migrate` i will get this error:
```
django.db.utils.IntegrityError: (1062, "Duplicate entry '2-change_config' for key 'auth_permission_content_type_id_codename_01a
b375a_uniq'")
```


camilonova:
Thank you

PR #283, opened 2018-05-07
Improve consistency of reset value handling for `date`

myii:
#### Summary

* Follow-on for #274 by @DmitryKaramin
* Consistent implementation re: `datetime`
* Consistent reset using `DATE_INPUT_FORMAT`, i.e.
  - Works whether `USE_L10N` setting is active or not

---

#### Explanation

I've been using v2.2.0 and had issues with handling dates.  Stated making my own modifications but then came across #274, which resolves the problem.

This PR makes the solution more consistent, both with the surrounding code as well as the actual reset value format, i.e. brings it into line with `datetime`.

### DIFF
--- a/constance/static/admin/js/constance.js
+++ b/constance/static/admin/js/constance.js
@@ -11,6 +11,9 @@
 
             if (fieldType === 'checkbox') {
                 field.prop('checked', this.dataset.default === 'true');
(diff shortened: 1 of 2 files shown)


camilonova:
Thank you.

myii:
You're welcome.

Issue #294, opened 2018-08-13
Django-constance not properly working with django 2.1


mitshel:
### Describe the problem
Command line manage command of Django-constance not properly working with django 2.1

### Steps to reproduce
`./manage.py constance list`

result is error:
```
Traceback (most recent call last):
  File "D:\Мои документы\Dropbox\Projects\Eclipse\workspace\sopds\manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\django\core\management\__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\django\core\management\__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\django\core\management\base.py", line 308, in run_from_argv
    parser = self.create_parser(argv[0], argv[1])
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\django\core\management\base.py", line 282, in create_parser
    self.add_arguments(parser)
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\constance\management\commands\constance.py", line 35, in add_arguments
    parser_list = subparsers.add_parser('list', cmd=self, help='list all Constance keys and their values')
  File "C:\Users\mitsh\AppData\Local\Programs\Python\Python37\lib\argparse.py", line 1116, in add_parser
    parser = self._parser_class(**kwargs)
  File "D:\Projects\Eclipse\workspace\sopds\venv\lib\site-packages\django\core\management\base.py", line 48, in __init__
    super().__init__(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'cmd'
```

### System configuration

* Django version: 2.1
* Python version: 3.5-3.7

jryannel:
Seems to be this line:
```
        parser_list = subparsers.add_parser('list', cmd=self, help='list all Constance keys and their values')
```

And following lines using cmd argument. Removing the "cmd" argument makes it work again.

I could not find a reference in ArgumentParser add_prser for the cmd parameter. Need to check the git log

rahulsoni11:
I'm Beginner in Django, 
Can I still be able to use Constance in my Django 2.1 Project, if yes How.?

mitshel:
[...]

PR #295, opened 2018-08-21
WIP: in django>=2.1 cmd was removed as parameter to add_parser

jryannel:
I have not tested it with django<2.1 so I can not tell which side-effects this may have. From the code only the error method seems to be potentially effected.

Ideally django should have popped "cmd" from the kwargs when changing the API in 2.1. For an analysis please see #294.

If this has side-effects another option would be to call two apis based on django version,e.g

```
if djang0.VERSION >= (2,1):
  # do one thing
else:
  # do another thing
```

### DIFF
--- a/constance/management/commands/constance.py
+++ b/constance/management/commands/constance.py
@@ -32,12 +32,12 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         subparsers = parser.add_subparsers(dest='command')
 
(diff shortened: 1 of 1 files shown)


camilonova:
Thank you, could you please squash your commits into one so we can merge it?

camilonova:
Thank you

PR #303, opened 2018-09-19
Add DummyBackend.

POD666:
It was useful for my case.
I will also add some notes to docs if you like it.


jezdez:
Can you elaborate what the use case for this is?

POD666:
In my case django-constance is used for staging environment only - it gives the ability to change settings for QA.
DummyBackend works faster in production.

camilonova:
I don't see this is a fit for the project.

I know there are ways to mock the redis connection for testing. I'll see if I can find it.

camilonova:
Thank you. I'll close this in favor of keeping the scope of the project as simple as possible.

PR #309, opened 2018-10-10
show not existing fields in field_list

horida:
When missing a field mentioned in `CONSTANCE_CONFIG_FIELDSETS` in `CONSTANCE_CONFIG`, the error message is no explicit about the variable.

`AssertionError: CONSTANCE_CONFIG_FIELDSETS contains field(s) that does not exist `

 To make the debugging easier, I include the missing variables in the error message

AssertionError: CONSTANCE_CONFIG_FIELDSETS contains field(s) that does not exist: example_field_1


camilonova:
Thank you!

PR #310, opened 2018-11-02
unittest for changelist_view and constanceform.save

zodman:


zodman:
close to use other PR

PR #313, opened 2018-12-27
Add field created_at and updated_at for model Constance

haos616:


camilonova:
I'm not sure why this is needed, can you elaborate more?

camilonova:
I don't think this is really needed. Thank you.

sabareesh-pixl:
"Having an updated_at field would be helpful to track when a particular setting was last modified. Alternatively, it would be ideal if the Constance model could be made customizable, allowing additional fields to be added from the application side. One approach could be to create an abstract base model with the necessary Constance functionality, which would give users more flexibility to extend it with custom fields."

PR #315, opened 2019-02-01
Allow null & blank for PickleField.

TkTech:
The default behaviour of PickleField was changed in `2.0` from `null=True` to
`null=False`. This causes issues with fields such as an image_field which
will try storing a `NULL` when unset. This small change is backwards and
forwards version safe.


camilonova:
Thank you.

PR #319, opened 2019-03-07
Use default_storage to save file

lemieux:
The current behavior of saving files is to save locally using `open` and `MEDIA_ROOT` which doesn't work when your storage is a remote one (eg. S3). Using `default_storage` will use whatever you had configured in your settings.


lemieux:
Tests seems to be failing on something unrelated to the changes (dependencies on Python 3.5).

therefromhere:
FYI @lemieux tests are passing on master now (fixed by #320)

camilonova:
Thank you

lemieux:
np! any info on when the next release will be cut?

PR #322, opened 2019-03-13
fixed "Reset to default" button with constants containing a space. closes #321

### LINKED ISSUE #321: "Reset to default" button doesn't work with constants whose name contains a space
### Describe the problem

When a constant's name contains a space, the "reset to default" action in the admin interface doesn't work. 

### Steps to reproduce

create a constant whose name contains a space

### System configuration

* Django version: 2
* Python version: 3

makerGeek:


camilonova:
Thank you

PR #325, opened 2019-03-28
made results table responsive for Django 2 admin

ilyachch:
Added wrapper like in Django 2 to make results table horizontal scrollable


camilonova:
Thank you

Issue #293, opened 2018-08-10
Unhelpful error message: CONSTANCE_CONFIG_FIELDSETS is missing field(s) that exists in CONSTANCE_CONFIG.


stitch:
### Describe the problem
If you have over a hundred fields defined in CONSTANCE_CONFIG, it's pretty evil to check them one by one to find which one you did not refer to in CONSTANCE_CONFIG_FIELDSETS. The big guestion is: what fields am i missing?

### Steps to reproduce

Create a few hundred fields and figure out which ones you used and which one you didn't.

### System configuration

* Django version:
* Python version:

camilonova:
Would you like to improve this library adding this check?

PR #326, opened 2019-04-05
Add a Django system check...

mozz100:
...that CONFIG_FIELDSETS accounts for all of CONFIG

In other words, adds a check that addresses the issue mentioned in https://github.com/jazzband/django-constance/issues/293


camilonova:
Great work. Thank you!

Issue #19, opened 2012-10-30
track the history of config changes


dineshkumardg:
Can you add a privilege to track the previous config changes???

aleszoulek:
Not sure if that should be in scope of the project. Maybe custom backend, that would be db-only and keeping that history?
Constance itself could just log the changes over python logging, so the user can monitor those in his logging system. (?)

MarZab:
Is there a specific reason `django_admin_log` can not be used?

camilonova:
Well, seems a good time to close this issue. I agree is not in the scope of the project.

PR #331, opened 2019-06-25
Feature/logentry

mavriq:
Hi, I added the ability to enable logging of constant changes.
I hope you find this useful.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -105,6 +105,9 @@ def get_values():
 class ConstanceForm(forms.Form):
     version = forms.CharField(widget=forms.HiddenInput)
 
+    class __ImpossibleException(Exception):
+        pass
+
     def __init__(self, initial, *args, **kwargs):
         super(ConstanceForm, self).__init__(*args, initial=initial, **kwargs)
         version_hash = hashlib.md5()
@@ -152,9 +155,13 @@ class ConstanceForm(forms.Form):
     def log_changed_field(self, name, old, new):
         if not settings.DATABASE_LOG_CHANGES:
             return
-        if not hasattr(self, '_request'):
-            return
-        if not hasattr(self, '_model_admin'):
+        try:
+            if not hasattr(self, '_request'):
+                raise self.__ImpossibleException('_request')
+            if not hasattr(self, '_model_admin'):
+                raise self.__ImpossibleException('_model_admin')
+        except self.__ImpossibleException as e:
+            ## TODO: log it, or use raven
             return
         #
         from constance.backends.database import DatabaseBackend
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -143,8 +143,36 @@ class ConstanceForm(forms.Form):
             self.cleaned_data[file_field] = file.name
 
         for name in settings.CONFIG:
-            if getattr(config, name) != self.cleaned_data[name]:
-                setattr(config, name, self.cleaned_data[name])
+            _current = getattr(config, name)
+            _new = self.cleaned_data[name]
+            if _current != _new:
+                setattr(config, name, _new)
+                self.log_changed_field(name, _current, _new)
+
+    def log_changed_field(self, name, old, new):
+        if not settings.DATABASE_LOG_CHANGES:
+            return
+        if not hasattr(self, '_request'):
+            return
+        if not hasattr(self, '_model_admin'):
+            return
+        #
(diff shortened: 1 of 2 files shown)


camilonova:
I'm not seeing this into the scope of this library.

mavriq:
In this case, tell me how best to do?
Should I fork this project? or can I enable logging of changes with the current library?
I see that the same request was in the past from other users #19

jezdez:
@mavriq Constance provides a [`config_updated` signal](https://django-constance.readthedocs.io/en/latest/#signals) that you can use to log changes. Signal receivers get both the old and the new value. The Django docs provide help how to [listen to signals](https://docs.djangoproject.com/en/2.2/topics/signals/#listening-to-signals).

I'm closing this PR for now since it's clearly a no-go.

Anton-Shutik:
@jezdez There are at least several people who would like to see such an option in the app. Some of them even provided a solution like this. Why not to merge it ? You can say: go ahead and fork, but I'm not a fun of doing this as well as a lot of other developers.

 @mavriq spent and "hour" making this PR, which might save this "hour" for hundreds of other developers who are using the app. Isn't it what opensource is great for ?

>Thank you. I'm not seeing this into the scope of this library.

@camilonova Totally disagree. That would be a great feature of this library.

PS.
Ready to help with implementing the feature.
Related issue https://github.com/jazzband/django-constance/issues/19

jezdez:
@Anton-Shutik The feature request was deemed out of scope of this app (read: the amount of time needed to maintain it outweighs the expected usefulness), so it doesn't matter if it was requested by many people, especially since there are alternatives.

The feature idea isn't bad of course, I mentioned that there is a relatively easy way to do this using the `config_updated` signal which is a low-maintenance-cost way to provide an official API to receive updates when a Constance config value changes.

Connecting a signal callback function to that signal to log the changes in a app-specific data model (or whatever logging mechanism is useful for your application) is simply left to the user of the app. This is a very common trade-off for edge case features in libraries such as Constance.

I'd like to encourage you to try implementing this in your project instead, or maybe write a blog post how to connect `config_updated` and the Django admin log facility. Good luck!

Anton-Shutik:
>The feature idea isn't bad of course, I mentioned that there is a relatively easy way to do this using the config_updated signal which is a low-maintenance-cost way to provide an official API to receive updates when a Constance config value changes.

Yep, that's why people would like to have that implemented as option in this app, rather then implementing in their project. Note that each dev have to implement it instead of being done once in the app and reused.

>I'd like to encourage you to try implementing this in your project instead, or maybe write a blog post how to connect config_updated and the Django admin log facility. Good luck!

That might be a good idea, but I think it won't have enough features to be considered as separate project, since it will closely related to constance.

jezdez:
> > The feature idea isn't bad of course, I mentioned that there is a relatively easy way to do this using the config_updated signal which is a low-maintenance-cost way to provide an official API to receive updates when a Constance config value changes.
> 
> Yep, that's why people would like to have that implemented as option in this app, rather then implementing in their project. Note that each dev have to implement it instead of being done once in the app and reused.

No, the existence of a developer API like a Django signal isn't a reason to implement a very specific feature in the library providing that API. That's just not what those things are there for. I know that you want this feature, but without investing time in maintaining it yourself. Which means we are at an impasse, having to agree to disagree.

> > I'd like to encourage you to try implementing this in your project instead, or maybe write a blog post how to connect config_updated and the Django admin log facility. Good luck!
> 
> That might be a good idea, but I think it won't have enough features to be considered as separate project, since it will closely related to constance.

You don't know until you try, a full-blown project may not be needed but instead a snippet somewhere, or simply a blog post titled "How to integrate django-constance with the Django admin to create an audit log".

Anton-Shutik:
I would not say that this is very specific feature. I can agree that a lot of (or even most of ) developers won't need it, but likely would use it if it exists. This is kinda "nice-to-have" feature. 

I didn't say that I want this feature without investing time in maintaining it myself. This is not true. I just said that once implemented it would save a bunch of time for other people.

I've added a [draft PR](https://github.com/jazzband/django-constance/pull/336), that adds the feature optionally. That might be helpful for other devs.

jezdez:
Thanks for opening the PR, this will be useful for others to implement it on their own. I've closed the PR as to not falsely suggest that it'll be merged. The feature continues to be out of scope of django-constance.

PR #333, opened 2019-07-09
ConstanceForm save method tweak to store the actual filename

riazanovslv:
We need to store the actual name of the file, which the storage's save method provides. Otherwise, if we upload a new file with the same name as the already stored file has, the Constance config will keep a link to the old file.


jmfederico:
Can this be merged please? 😃

camilonova:
Thank you.

PR #335, opened 2019-07-19
use return value of default_storage.save() as path

jmfederico:
`default_storage.save()` returns the final path where the file was saved too. It may be different from the requested path.

According to Django's documentation on [`default_storage.save()`](https://docs.djangoproject.com/en/2.2/ref/files/storage/#django.core.files.storage.Storage.save) **...If there already exists a file with this name name, the storage system may modify the filename as necessary to get a unique name...**

Therefore It is necessary to use the returned path as the value for the Constance config.


jmfederico:
Well, there is another PR with the same change. #333

PR #338, opened 2019-08-11
Allow override_config for pytest

TamoshaytisV:
`override_config` is currently possible to apply only to unittest TestCase classes.
This PR adds `constance.test.pytest` module that can be used as follows

Examples can be found in [doc](https://github.com/TamoshaytisV/django-constance/blob/feature/pytest_override_config/docs/testing.rst#pytest-usage)

It's still possible to do import as usual for unittest
```py
from constance.test import override_config

...
```


constance/test/unittest.py
```diff
@@ -0,0 +1,25 @@
...
+    itself.
+    """
+    _pre_setup = "_pre_setup"
+    _post_teardown = "_post_teardown"
+
+    def __call__(self, test_func):
+        if isinstance(test_func, type):
+            if not issubclass(test_func, SimpleTestCase):
+                raise Exception(
+                    "Only subclasses of Django SimpleTestCase can be "
```
TamoshaytisV: Possibly, in this error, we can let a user know about alternative. But i think it's better to update the docs only

camilonova:
Vladas thank you for your contribution. Why this is needed?

TamoshaytisV:
There are a lot of django projects where i use pytest as a test runner. Django-constance forces me to use TestCase inheritance to make override work
I have the workaround to set `CONSTANCE_CONFIG` values it `test.py` settings but often i want to have different overrides for specific methods or classes

@camilonova please let me know if it's clear enough

camilonova:
Got it. I'm not able to make a decision here. Maybe @jezdez knows best.

constance/test/base.py
```diff
@@ -37,27 +30,9 @@ def inner(*args, **kwargs):
...
     def modify_test_case(self, test_case):
         """
-        Override the config by modifying TestCase methods.
-
-        This method follows the Django <= 1.6 method of overriding the
-        _pre_setup and _post_teardown hooks rather than modifying the TestCase
-        itself.
+        Override the config by modifying TestClass methods.
         """
-        original_pre_setup = test_case._pre_setup
```
TamoshaytisV: moved to `constance.test.unittest`

constance/test/base.py
```diff
@@ -1,32 +1,25 @@
...
+        super(OverrideConfigBase, self).__init__(**kwargs)
         self.original_values = {}
 
     def __call__(self, test_func):
         """
-        Modify the decorated function to override config values.
+        Modify the decorated method/class to override config values.
         """
         if isinstance(test_func, type):
-            if not issubclass(test_func, SimpleTestCase):
```
TamoshaytisV: moved to `constance.test.unittest`

jezdez:
Thanks @camilonova for raising this with me, I agree that this is a non-trivial feature that we need to discuss first.

In general I'm in favor of improving the use of override_config when used with Pytest. But I'm not sure if providing a decorator like override_config fits the best practices of the Pytest community.

E.g. pytest-django didn't implement a custom overrride_settings decorator to override Django's settings but instead chose to implement [a Pytest fixture that allows modifying of settings](https://github.com/pytest-dev/pytest-django/blob/c1bdb8d61498f472f27b4233ea50ffa2bce42147/pytest_django/fixtures.py#L305-L345) in the test session.

Similarly I would encourage you to implement this feature as a `constance` or `config` fixture that would have the ability to override config value during individual tests. I would consider that to be a more practical user experience than having to know which decorator to use.

TamoshaytisV:
@jezdez I agree that for pytest it's better to use fixtures or markers but what initially stopped me from implementing in that way is complexity to make it work for any scope.
I will investigate more using pytest-django example and will get back with results soon.
Thank you

jezdez:
> @jezdez I agree that for pytest it's better to use fixtures or markers but what initially stopped me from implementing in that way is complexity to make it work preserving simplicity of decorator and ability to change config for any scope.

I understand, and appreciate your openness to investigate this further. As always I try to apply the mantra "it's easy to add, but is it also easy to maintain?" to such discussion. Your input and contribution is valued and welcome nevertheless.

> I will investigate more using pytest-django example and will get back with results soon.
> Thank you

TamoshaytisV:
@jezdez @camilonova sorry for being inactive for a long time. I finalized refactoring and feature itself, please review or feel free to assign propper reviewers if you are not available.

constance/test/pytest.py
```diff
@@ -0,0 +1,79 @@
...
+            pytest.fail(
+                "Constance override can not not accept positional args"
+            )
+        with ConstanceConfigWrapper(**marker.kwargs):
+            yield
+    else:
+        yield
+
+
+class ConstanceConfigWrapper(object):
```
TamoshaytisV: Probably the need to have `__setattr__` in this class may appear but ATM i can't think of many usage cases and all they can be solved using context manager or uning fixture and marking tests/classes/modules

Mogost: Since there is no more support for Python 2, there is no need for inheritance from object
```suggestion
class ConstanceConfigWrapper:
```

docs/testing.rst
```diff
@@ -38,3 +38,67 @@ method level and also as a
...
+        ...
+
+Class/function scope
+
+.. code-block:: python
+
+    from constance import config
+
+    @pytest.mark.override_config(API_URL="/awesome/url/")
+    class SomeClassTest(object):
```
Mogost:
```suggestion
    class SomeClassTest:
```

docs/testing.rst
```diff
@@ -38,3 +38,67 @@ method level and also as a
...
+
+Pytest usage
+~~~~~
+
+Django-constance provides pytest plugin that adds marker
+:class:`@pytest.mark.override_config()`. It handles config override for
+module/class/function, and automatically revert any changes made to the
+constance config values when test is completed.
+
+.. py:function:: pytest.mark.override_config((**kwargs)
```
Mogost: Something is wrong with brackets

TamoshaytisV: Good catch! Fixed

docs/testing.rst
TamoshaytisV: Thanks! Changed

constance/test/pytest.py
TamoshaytisV: You are right, changed

marco-silva0000:
I like what's been done here, I'm using a copy paste on my code and it makes tests really clean.
+1 for this to be merged in

TamoshaytisV:
@jezdez will you be so kind to assign other reviewers if you are busy with smth else?

constance/test/pytest.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __enter__(self):
+        self.enable()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.disable()
+
+
+@pytest.fixture
+def override_config():
```
pawelzar: Maybe it would be better if this was just called `config`? Similarly to `settings` fixture for Django settings. `override_config` better suits as a name for decorator/context manager.

constance/test/pytest.py
```diff
@@ -0,0 +1,79 @@
...
+            pytest.fail(
+                "Constance override can not not accept positional args"
+            )
+        with ConstanceConfigWrapper(**marker.kwargs):
+            yield
+    else:
+        yield
+
+
+class ConstanceConfigWrapper:
```
pawelzar: You could use `ContextDecorator` which would allow to use it as decorator and context manager:
```suggestion
class override_config(ContextDecorator):
```
https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator

constance/test/pytest.py
camilonova: I agree

camilonova:
Please check the latest changes/comments and I would be happy to merge this.

Thank you all!

TamoshaytisV:
@pawelzar I used `ContextDecorator` so now it's possible to use `override_config` in the same way as it is for unittest. Nevertheless, I insist to keep fixture and mark so users can use it in pytest way but with less code

camilonova:
Thank you all

utapyngo:
Is this going to be released soon?

camilonova:
If someone can confirm it works fine if the package is installed from master I can make a release

TamoshaytisV:
I've tried it in my project with
```bash
-e git+https://github.com/jazzband/django-constance.git#egg=constance[database]
```
I've got warning
`WARNING: Generating metadata for package constance produced metadata for project name django-constance. Fix your #egg=constance fragments.`

In fact, lib was really installed with database backend anyway, so in general, it's working
Can someone please correct my `egg=` part and also try it as I'm not sure what's the correct way to specify `extras_require` for github url?

constance/test/pytest.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+    )
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_call(item):  # pragma: no cover
+    """
+    Validate constance override marker params. Run test with overrided config.
+    """
+    marker = item.get_closest_marker("override_config")
```
Mogost, line 30: Need something like [here (in pytest-django)](https://github.com/pytest-dev/pytest-django/blob/ee7858af1a80e0091af4d260c0a70e5766c1196a/pytest_django/plugin.py#L421) to prevent #410 
```suggestion
    if not django_settings_is_configured():
        return
    marker = item.get_closest_marker("override_config")
```

PR #345, opened 2019-10-05
Updated checks for additional detail and fixes

dstrohl:
Added checks for no constance config, missing items, and dupe items.

Also, the checks were not working for me, (not sure if that was because they were checking for "CONFIG" instead of "CONSTANCE_CONFIG", or if there was another reason), but this version does seen to work.


dstrohl:
closed... oops, I realized that CONSTANCE_CONFIG is being renamed CONFIG.... I'll fix my pull request, but I'm not sure why now your check didnt work.

PR #346, opened 2019-10-05
Trying pull request again... added additional checks.

dstrohl:
For some reason the initial one did not work on my system, this one did.  but along the way I added a few more checks to check for 

Errors:
- missing items in fieldssets
- extra items in field-sets

Warnings:
- constance not configured at all
- duplicate items in field-sets. (perhaps this should be an error, but I wasn't sure if you expected some people to put something under two sections) and how your logic would handle it.

### DIFF
--- a/constance/checks.py
+++ b/constance/checks.py
@@ -1,3 +1,4 @@
+
 from django.conf import settings
 from django.core import checks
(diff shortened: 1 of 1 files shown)


camilonova:
Please fix the Travis build. Thank you.

Issue #360, opened 2019-12-17
[Improvement] TypeError: 'Config' object is not subscriptable


bartmachielsen:
### Describe the problem

It would be a usefull feature if you can dynamically call the variables in constance. Small addition by adding the method `def __getitem__(self, key):` to the Config class.

### Steps to reproduce
Don't use the `config.SOME_CONSTANCE_VARIABLE` method but call by string.
`config['SOME_CONSTANCE_VARIABLE']`

Work around is to call the get_attr function directly (but not the most prettiest option)

camilonova:
@jezdez Do you think is something we should do?

I'm not sure if we should allow Constance to work as a dict.

Mogost:
This seems like a reasonable change, but if it is accepted, documentation must be added.
I would start the discussion by answering the question “why not?”.

jezdez:
We shouldn't make the Config class a hybrid container since it'd would just be a matter of time till people wonder why it doesn't have all the dict behaviors. It's perfectly valid to use `getattr` for this kind of use case.

PR #361, opened 2019-12-17
get_item and set_item impl

bartmachielsen:
See #360 for details about request

### DIFF
--- a/constance/base.py
+++ b/constance/base.py
@@ -9,6 +9,12 @@ class Config(object):
         super(Config, self).__setattr__('_backend',
             utils.import_module_attr(settings.BACKEND)())
 
+    def __getitem__(self, key):
+        return self.__getattr__(key)
+    
+    def __setitem__(self, key, value):
+        return self.__setattr__(key, value)
+
     def __getattr__(self, key):
         try:
             if not len(settings.CONFIG[key]) in (2, 3):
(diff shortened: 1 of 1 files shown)


Mogost:
I consider this change acceptable. But in order for this PR to be merged, it is necessary to document this feature.

constance/base.py
```diff
@@ -9,6 +9,12 @@ def __init__(self):
         super(Config, self).__setattr__('_backend',
             utils.import_module_attr(settings.BACKEND)())
 
+    def __getitem__(self, key):
+        return self.__getattr__(key)
+    
+    def __setitem__(self, key, value):
+        return self.__setattr__(key, value)
+
```
Mogost: In my opinion, this variation looks simpler.
```python
    __getitem__ = __getattr__
    __setitem__ = __setattr__
```

Mogost: @bartmachielsen This must be declared after the associated methods are defined. Like that
```python
    def __getattr__(...):
        ...

    __getitem__ = __getattr__

    def __setattr__(...):
        ...

    __setitem__ = __setattr__
```

constance/base.py
```diff
@@ -9,6 +9,9 @@ def __init__(self):
         super(Config, self).__setattr__('_backend',
             utils.import_module_attr(settings.BACKEND)())
 
+    __getitem__ = __getattr__
```
tayyab-razzaq: it should be done after the `__getattr__` method definition

jezdez:
Following closing the issue in https://github.com/jazzband/django-constance/issues/360#issuecomment-573616410 this PR isn't needed anymore.

PR #371, opened 2020-01-30
delete south migrations

syurchenko:
South migrations can be deleted because support of old Django versions is dropped.


camilonova:
Thank you

PR #378, opened 2020-02-12
Simplify documentation installation section.

### LINKED ISSUE #375: ImproperlyConfigured: The Redis backend requires redis-py to be installed.
### Describe the problem

The following exception is raised: 
`ImproperlyConfigured: The Redis backend requires redis-py to be installed.`

### Steps to reproduce

- Run `pip install "django-constance[database]"`
- Add `constance` to `settings.INSTALLED_APPS`:
```python
INSTALLED_APPS = (
    'django.contrib.admin',
    'django.contrib.staticfiles',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    ...
    'constance',
)
```
- Navigate to `/admin/constance/config/`

### System configuration

* Django version: `3.0.3`
* Python version: `3.7.3`

martey:
Make "Installation" section of documentation less fraught by only including instructions to install django-constance with the default Redis backend. This will reduce the chances of new users failing to
successfully install the database backend because it needs additional configuration.

Fixes #375.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,23 +9,13 @@ Features
 
 .. image:: screenshot2.png
 
-Installation
-------------
-
-Install from PyPI the backend specific variant of django-constance:
-
-For the (default) Redis backend::
+Quick Installation
+------------------
 
     pip install "django-constance[redis]"
 
-For the database backend::
-
-    pip install "django-constance[database]"
-
-Alternatively -- if you're sure that the dependencies are already
-installed -- you can also run::
-
-    pip install django-constance
+For complete installation instructions, including how to install the
+database backend, see :ref:`Backends <backends>`.
 
 Configuration
 -------------


docs/index.rst
```diff
@@ -9,24 +9,11 @@ Features
...
 
-Install from PyPI the backend specific variant of django-constance:
-
-For the (default) Redis backend::
-
     pip install "django-constance[redis]"
 
-For the database backend::
-
-    pip install "django-constance[database]"
```
Mogost: You have completely excluded information about `database` backend. This is not a good approach.
Probably there really is some kind of problem with the documentation in this aspect, but just not mentioning the `database` backend is not a solution.

rmaciejczyk: I personally use the `database` backend at work and when I started I was looking for some configuration instructions. I believe there are many more users who also need it. So I agree with @Mogost it definitely shouldn't be removed from the docs.

martey: I also use the `database` backend. I think there's some confusion here - *this pull request doesn't completely remove information about the `database` backend from the documentation*, it removes *incomplete* information that causes people trying to use the `database` backend to run into problems.

I think adding a link to the Backends page (which contains a complete set of installation instructions for the `database` backend) to the Installation section should assuage the concerns raised here.

Technically, updating the Installation section to include complete instructions would also accomplish the same goal. However, since this would duplicate the information on the Backends page, I don't think that would be the best solution.

PR #329, opened 2019-06-05
Optimizations for database backend

xuhcc:
Reduced number of DB queries:
* get default: from 6 to 3
* set: from 3 to 2

PR #384, opened 2020-03-05
Allow concurrent calls to set() method

### LINKED ISSUE #383: django.db.utils.IntegrityError after 2.5.0 release
In #329 `DatabaseBackend.set` was changed. And now I have `IntegrityError` in my project.

The bottom line is that when doing parallel requests, the following happens:
On execute 
https://github.com/jazzband/django-constance/blob/50287a3b2a3096ea89f6c28ba11935949fe21a27/constance/backends/database/__init__.py#L87
raised `self._model.DoesNotExist`
Then executed
https://github.com/jazzband/django-constance/blob/50287a3b2a3096ea89f6c28ba11935949fe21a27/constance/backends/database/__init__.py#L93
raises `django.db.utils.IntegrityError` due to the fact that another thread was able to set the value earlier.

I cannot provide a test that will demonstrate this error.
I call here PR author @xuhcc, merger @camilonova and @jezdez as an observer.

xuhcc:
It turned out that changes made in #329 led to errors when `set()` method is called simultaneously from different processes. This PR fixes #383 by replicating the logic of `get_or_create()` which previously used in this part of code. Setting the default value now requires 5 database queries which is still an improvement if we compare with the time before #329 when it took 6 queries.

### DIFF
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -1,7 +1,12 @@
 from django.core.cache import caches
 from django.core.cache.backends.locmem import LocMemCache
 from django.core.exceptions import ImproperlyConfigured
-from django.db import OperationalError, ProgrammingError
+from django.db import (
+    IntegrityError,
+    OperationalError,
+    ProgrammingError,
+    transaction,
+)
 from django.db.models.signals import post_save
 
 from .. import Backend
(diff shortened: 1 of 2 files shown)


Mogost:
I can confirm that this code fix my broken tests.
But we should pay more attention to the review of this code.

constance/backends/database/__init__.py
```diff
@@ -82,19 +87,32 @@ def get(self, key):
...
+            try:
+                with transaction.atomic(using=queryset.db):
+                    queryset.create(key=key, value=value)
+                created = True
+            except IntegrityError as error:
+                # Allow concurrent writes
+                constance = queryset.get(key=key)
+
+        if not created:
             old_value = constance.value
```
Mogost: **_From PyCharm:_** Local variable `constance` might be referenced before assignment

Mogost:
@camilonova Please make a new release.

PR #390, opened 2020-04-21
Russian locale update

crazyzubr:


camilonova:
I'm not sure we want to have custom labels as part of the library. What are your motivations for that?

crazyzubr:
@camilonova for example, i'm want to add  a parameter **robots.txt**. I will have to put such a parameter name _robots_txt_, because in **robots.txt** contain dot. So auto_id for input generated  id_robots.txt. After which it is impossible to press the button "Reset to default", because this id is incorrect in DOM. The label can contain any characters: dots, spaces, etc.

camilonova:
@crazyzubr Thank you for your time. I think this is a ver specific case when you need to use special characters at the name. You can get by using the underscore instead of the dot. Think as if this is a variable name where you are limited on how can you name it.

This change will add more complexity to the base code so I will close this.

Issue #254, opened 2018-02-15
Attempts to change config values fail silently and appear to succeed when user does not have change permissions


kerkeslager:
### Describe the problem

Users with `is_superuser` set to `False` can view constance configs, but not edit them, by default. The constance admin interface UI shows no indication that the fields are readonly, or that the user does not have permission to edit them, and even makes it seem like the user has edited the field when they try to save changes.

This took a while to debug because it fails completely silently with no error message. As such I feel it could be friendlier.

Additionally, I'm not sure that letting non-superuser users view constance config values is a sensible default.

The issue in my case was solved by giving the staff user permissions, but this was found only on second reading of the documentation, as there's only a few lines that indicate anything about permissions in the docs, and no messaging in the admin interface indicates why attempts to change the constance configs failed.

### Steps to reproduce

1. Install constance
2. Access the admin interface as a user who a) is not a superuser, and b) does not have constance edit permissions.
3. View the constance config in the admin interface.
4. Attempt to edit the constance config (it will appear that the constance config has changed).
5. Go to a different area of the admin site and return to the constance configs.
6. At this point it will be apparent that the config has not changed.

### System configuration

* Django version: 1.8.18
* Python version: 2.7.14
* Django constance version: 2.1.0

Allan-Nava:
Is still alive this project?

jezdez:
@Allan-Nava Hi there, I see you've asked the same question in https://github.com/EliotBerriot/django-dynamic-preferences/issues/136 and https://github.com/koralarts/django-dynamic-settings/issues/4 which made me wonder what you're trying to achieve by asking it. This project had a commit 3 days ago, so it seems moot to ask if it's alive.

PR #393, opened 2020-04-26
Read-only form with correct perm

SebCorbin:
Ref #254 

When `CONSTANCE_SUPERUSER_ONLY` is set to False, there is no way of displaying the values without the change permission.

This PR fixes this issue with the new `view_config` permission.


camilonova:
Thank you

Issue #349, opened 2019-10-31
Provide a 'setting_defaults' backend for testing


RichardForshaw:
### Describe the problem

I have an issue/request that is similar to #285. In production and integration testing I run a redis backend, but in an effort to minimize complexity when running unit tests I want to be able to minimize the additional services that are running, such as redis.

### Suggested improvement

My suggestion is to have some kind of 'defaults' backend, which will simply return the default values defined in the `CONSTANCE_CONFIG` structure, unless they are overridden with an `override_config` directive. I suspect that this is reasonable behaviour when writing unit tests, as you will tend to use setting defaults unless you specifically want to exercise the setting values.

### Steps to reproduce

This is a request rather than an issue. The current behaviour is standard.

### System configuration

* Django version: N/A
* Python version: N/A

bmwant:
@RichardForshaw check out this PR https://github.com/jazzband/django-constance/pull/394

camilonova:
This will be addressed in #394

Issue #285, opened 2018-05-16
Using Local-memory as Redis replacement in test environment


Jastrzebowski:
### Describe the problem

I'm using Redis as a backend for both caching and Constance, in my test env I'm replacing Redis with Local-memory caching. How should I use Constance in this setup?

### System configuration

* Django version: 2.0.5
* Python version: 3.6.5

belyak:
Really important. At the moment I work on the system with sophisticated data processing and it is very uncomfortable to restore local Constance settings every time after the unit-tests run (I have to use and unit and manual testing locally)
As a workaround you can so the next:
(in settings.py)

```
TEST_MODE = 'test' in sys.argv

if TEST_MODE:
    # change constance DB to preserve existing local settings:
    CONSTANCE_REDIS_CONNECTION = 'redis://localhost:6379/1'


```

bmwant:
See this PR https://github.com/jazzband/django-constance/pull/394

Mogost:
Done in #394

PR #394, opened 2020-05-09
Add simple backend

bmwant:
### Rationale

Provide in-memory backend for usage in unittests when no external service is available as per https://github.com/jazzband/django-constance/issues/285. 

It can also be used to cover this https://github.com/jazzband/django-constance/issues/349 use case.

### Usage

Update your test config in the app which uses `django-constance` library with 

```
CONSTANCE_BACKEND = 'constance.backends.simple.SimpleBackend'
```

Run your tests as usually (e.g. `python manage.py test`) using this in-memory backend without external database / redis instance.

### Changes

* Added new `SimpleBackend` in-memory backend
* Moved backend tests to `tests/backends` subdirectory
* Added tests for `mget` backend method

### TODO

* [ ] Update documentation if PR is going to be merged


bmwant:
@jezdez @camilonova could you review this?

camilonova:
Looks good to me. Please add documentation so we can merge this one 🍺

jezdez:
Hey, I haven't looked in detail into the implementation, but I'm thinking we should name this "MemoryBackend" or something similar since "simple" is rarely descriptive enough for users and instead could mislead them about its use case.

Secondly since the main purpose of the backend is to be used for testing, can we make sure that this is thread-safe when accessing the underlying dict?

bmwant:
@jezdez I've refactored code to `MemoryBackend` and added lock for thread safety. Python built-in structures and their methods are thread-safe by default, so the only place where race condition can occur is when retrieving multiple values (context switch while other thread modifying a setting), so I've modified `mget` method to reflect that.

constance/backends/memory.py
```diff
@@ -0,0 +1,35 @@
...
+from .. import signals, config
+
+
+class MemoryBackend(Backend):
+    """
+    Simple in-memory backend that should be mostly used for testing purposes
+    """
+    def __init__(self):
+        super().__init__()
+        self._storage = {}
```
jezdez, line 13: I'm probably a bit uncertain about threading internals, but I think we should move this into a module global so separate memory backend instances in the individual threads are using the same global storage, since that's the use case of constance.

bmwant: ok, I've got your point, will make corresponding updates

bmwant:
@jezdez Django uses locks within `loclmem` because its methods are not atomic (there are a couple of operations over the storage), but I agree we should stay close to that implementation.

I have also tested current solution with
```
def test_same_storage():
    s1 = MemoryBackend()
    s2 = MemoryBackend()
    assert id(s1._storage) == id(s2._storage)

    s1.set('s1-key', 's1-value')
    s2.set('s2-key', 's2-value')

    assert s1.get('s2-key') == s2.get('s2-key')
    assert s2.get('s1-key') == s1.get('s1-key')

    assert s1._storage == s2._storage


def test_same_lock():
    s1 = MemoryBackend()
    s2 = MemoryBackend()
    assert id(s1._lock) == id(s2._lock)
```

and with 

```
from threading import Thread

from constance.backends.memory import MemoryBackend

def target(target_name):
    s = MemoryBackend()
    s.set(target_name, 'value')

    print(s.get(target_name))
    print(id(s._storage))

def test_threads_share_storage():
    s = MemoryBackend()
    t1 = Thread(name='t1', target=target,  args=('t1',))
    t2 = Thread(name='t2', target=target, args=('t2',))

    t1.start()
    t2.start()

    t1.join()
    t2.join()
    # check storage in main thread
    assert len(s._storage) == 2
    assert s.get('t1') == 'value'
    assert s.get('t2') == 'value'
```

to make sure backend works properly and it's equivalent to have `_storage` and `_lock` on class level and on the module level.

PR #397, opened 2020-05-25
Normalize newlines from form

### LINKED ISSUE #396: Wrong "is modified" in admin for multi line strings
My default value for variable looks like 'Some\n\nText'
After this variable is init, saved and desirialized that looks like 'Some\r\n\r\nText'
For user in admin that looks strange.
Problem line is:
https://github.com/jazzband/django-constance/blob/731514c52fe1e669f6210a8a860bff7846c6f707/constance/admin.py#L208
How about use that approach to compare strings?
```
 'modified': localize(value).splitlines() != localize(default).splitlines()
```

Mogost:
Close #396

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -19,6 +19,7 @@ from django.utils import timezone
 from django.utils.encoding import smart_bytes
 from django.utils.formats import localize
 from django.utils.module_loading import import_string
+from django.utils.text import normalize_newlines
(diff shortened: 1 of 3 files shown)


Mogost:
Thanks @utapyngo for the tests

Mogost:
@camilonova @jezdez Could you take a look at that?

PR #402, opened 2020-06-05
Add anchors in admin

Mogost:
Very often there is a need to provide links to a specific setting, anchors would be very useful.
image


jezdez:
I really like the idea but visually this can be improved, maybe by just using the item name as the link text instead of adding the extra character.

Mogost:
@jezdez 
I originally did so, but I didn’t like that the whole title behaves like a link. This is reflected in colors, etc. In general, ideally, we need to add some more css so that the symbol is shown exclusively when you hover over the inscription. An example of this behavior is in github.
image

jezdez:
Yeah, fair point, let's do the hover idea.

Mogost:
@jezdez done

jezdez:
Could you add a new screenshot please?

jezdez:
Or has this not changed visually other than the hover effect?

Mogost:
image
on hover
image
I tried this on my project, at first glance it works well and copes equally well with both long and short variable names.
In the worst case scenario, if I did not take something into account, this will not lead to critical consequences and we will be informed about this in issues.

Mogost:
@jezdez I have also been waiting for a new release for a very long time. I'm still using 2.4.0 in my project because #383 affected me.

PR #418, opened 2020-08-21
url() is deprecated in Django 3.1

ekinertac:
replaced url() functions with newer re_path() function and add path() functions to urls.py files


jairhenrique (approved):

camilonova:
Thank you

PR #420, opened 2020-09-24
Persian (fa) language translation files

mavenium:
Hi,
I translated the translation files into Persian (fa).


camilonova:
Thank you

PR #424, opened 2020-10-14
fixed field_list bug in checks.py

mehrdel2002-hub:
fixed field_list bug, that get CONSTANCE_CONFIG_FIELDSETS tuple instead of get CONSTANCE_CONFIG_FIELDSETS fields dicts


Mogost (changes requested):
Something went wrong and the failed tests confirm it.

PR #425, opened 2020-10-17
fixed field_list bug

mehrdel2002-hub:
fixed field_list bug, that get CONSTANCE_CONFIG_FIELDSETS tuple instead of get CONSTANCE_CONFIG_FIELDSETS fields dicts


camilonova:
Looks like it was already fixed in ce37f630f139cf4a71f60a5af095072b03f5b259

PR #433, opened 2020-11-17
Fix for the get_inconsistent_fieldnames() function. The fieldset can be a dictionary, not a tuple.

mandalay-rp:
The fix given the possibility that the fieldset could be a dictionary (when it defined as collapsible).
https://django-constance.readthedocs.io/en/latest/#fieldsets-collapsing
```
CONSTANCE_CONFIG = {
    'SITE_NAME': ('My Title', 'Website title'),
    'SITE_DESCRIPTION': ('', 'Website description'),
    'THEME': ('light-blue', 'Website theme'),
}

CONSTANCE_CONFIG_FIELDSETS = {
    'General Options': {
        'fields': ('SITE_NAME', 'SITE_DESCRIPTION'),
        'collapse': True
    },
    'Theme Options': ('THEME',),
}
```

Currently, if the set is defined as collapsible, there is no way to save changes due to incorrect validation.

### DIFF
--- a/constance/checks.py
+++ b/constance/checks.py
@@ -2,7 +2,6 @@ from django.core import checks
 from django.utils.translation import ugettext_lazy as _
 
 
-
 @checks.register("constance")
 def check_fieldsets(*args, **kwargs):
     """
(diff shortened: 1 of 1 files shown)


camilonova:
Thank you

PR #436, opened 2020-11-30
Remove unnecessary post_save cachebust signal

jarekwg:
Hiyas,

I believe this code has been superseded by https://github.com/jazzband/django-constance/commit/1b6730d2c59d7e90d9eaad77d86b110de726f54f#diff-46b5df3b78d9475f4f60645d8fcd7cf897b6f789bae6c1d8782c105c34d714e3R83-R84

The code linked above updates the cache for the given key that has just been set.
The code I'm proposing to be removed updates the cache across _ALL_ keys whenever a single key is saved.

Not only is this inefficient, but it causes an undesirable interaction when changing constance values from within transactions (this is the reason I'm bringing it up).
Basic example:
- Key `A` is set to `4`.
- I increment key `A` from within a transaction. It takes on a value of `5`. Cache updates accordingly.
- Attempting to set key `A` from outside this transaction errors, as is to be expected; the row is locked until the transaction exits.
- Attempting to set key `B` from outside this transaction is allowed, as `B` is a different row in the db. This action sets `B` correctly, but has a sideeffect of reloading the _entire_ cache, putting `A` back to `4`.
- I read key `A` from within my transaction -- it's now `4` again; should be `5`.

Annoyingly, there appear to be no tests covering caching at all, as I can just as easily remove the other cache-setting section, and they'll all still pass.. I guess I can offer to add some, if this change is deemed acceptable?

Cheers,

### DIFF
--- a/constance/backends/database/__init__.py
+++ b/constance/backends/database/__init__.py
@@ -7,7 +7,6 @@ from django.db import (
     ProgrammingError,
     transaction,
 )
-from django.db.models.signals import post_save
 
 from .. import Backend
 from ... import settings, signals, config
@@ -37,8 +36,6 @@ class DatabaseBackend(Backend):
         else:
             self._cache = None
         self.autofill()
-        # Clear simple cache.
-        post_save.connect(self.clear, sender=self._model)
 
     def add_prefix(self, key):
         return "%s%s" % (self._prefix, key)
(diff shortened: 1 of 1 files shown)


camilonova:
Thank you. It seems make sense. Can you add some tests to ensure this is the way we want signals to behave?

PR #438, opened 2020-11-30
add arabic translation

medbenmakhlouf:


jezdez:
This needs to be compiled to a mo file

medbenmakhlouf:
> This needs to be compiled to a mo file

Just added that. Thanks

Issue #348, opened 2019-10-23
Unwanted reset of parameter to default value


lada-olc:
### Describe the problem

Time to time, I experience resetting of one parameter to the default value (False to True) and therefore enabling certain feature of our system.

### Steps to reproduce

By logging to Sentry, I located the problem in the database backend of django-constance. See the comments in Config.__getattr__ of base.py:
```python
    def __getattr__(self, key):
        try:
            if not len(settings.CONFIG[key]) in (2, 3):
                raise AttributeError(key)
            default = settings.CONFIG[key][0]
        except KeyError:
            raise AttributeError(key)

        # here I get None even tough the parameter is in the database:
        result = self._backend.get(key)
        if result is None:
            result = default

            # so the default value is set:
            setattr(self, key, default)
            return result
        return result
```

How the DatabaseBackend.get method could return None:
```python
    def get(self, key):
        key = self.add_prefix(key)
        if self._cache:
            value = self._cache.get(key)
            if value is None:
                self.autofill()
                value = self._cache.get(key)
        else:
            value = None
        if value is None:
            try:
                value = self._model._default_manager.get(key=key).value
            except (OperationalError, ProgrammingError, self._model.DoesNotExist):
                # HERE it is
                pass
            else:
                if self._cache:
                    self._cache.add(key, value)
        return value
```

I put logging command before the line with pass to get the specific Exception information next time, but still - is it correct to set the value to the default value when reading from database fails?

### System configuration

* Django version: 2.1.5
* Python version: 3.7.1

auuron:
Are there any plans to fix that?

rmaciejczyk:
I'm seeing pretty similar issue that some constance settings are sometimes reset to default values. What is weird I can see the correct values in the constance Django admin form but when I try to get them in the code they are default. I didn't debug it but it looks like this is the same issue.

jack-monk:
I encountered same issue on our production stack, we sometimes get temporary connection issue between VMs where neither DB VM nor memcache VM is available. Problem is hard to catch, as exceptions are silenced and have big impact cause it permanently changes DB fields values to defaults which are development values. We are forced to use production values as defaults instead of relying on values from DB, making django-constance in this case useless :( 
Are there plans to rework this logic?

dabrowne:
FYI for anyone still dealing with this issue. We mitigated it by overriding the default DatabaseBackend. `OperationalError` and `ProgrammingError` now fail loudly rather than silent fallback to default value.

```python
[...]

PR #444, opened 2021-03-03
Don't call backend.set when backend.get returns None and default value is None

i2xS:
Hi there!

**Preconditions:** 
Any constance variable 
default value: None
current value: None

Reading value through config.VALUE_NAME leads to backend.set('VALUE_NAME', None) call, which updates None to None.

This behaviour led us to database failure on production when 80 simultaneous value reads happen, which eventually led to  simultaneous writes and deadlocks. Deadlocks led to service restarts, emitting new connections to database and new deadlocks.

### DIFF
--- a/constance/base.py
+++ b/constance/base.py
@@ -17,7 +17,7 @@ class Config:
         except KeyError:
             raise AttributeError(key)
         result = self._backend.get(key)
-        if result is None:
+        if result is None and default is not None:
             result = default
             setattr(self, key, default)
             return result


Mogost (approved):

jezdez (changes requested):
Please provide a test case to make sure this won't break again.

camilonova:
@i2xS should we close this one?

MichaelDanielTom:
Any reason not to merge this? Seems like this should be the desired behavior, and it's still relevant given [this](https://github.com/jazzband/django-constance/issues/348).

camilonova:
@i2xS @MichaelDanielTom can you provide a test so we can merge this?

PR #446, opened 2021-03-15
Update __init__.py

Mogost:


jezdez:
I noticed that the version matrix is pretty hard to understand, could we refactor that a bit?

jezdez:
This should suffice at the moment:

```
    py{36,37,38,39,py3}-dj{22,30,31,32}-{unittest,pytest}
    py{38,39}-djmain-{unittest,pytest}
```

jezdez:
That's following the list on https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

Fazel94:
Hi, 
Could you explain the purpose of the `import django` here? and checking of if the django is less than 3.2 to define app, can it cause import errors? a fellow developer is facing occasionally import errors [here](https://stackoverflow.com/q/68962519/2099608) I suggested to use this version but couldn't come up with proper reasoning why would this help.

PR #448, opened 2021-04-08
Add a note about working around the locmem cache restriction

jacobian:
Per a conversation on twitter with @jezdez about this, here's a stab at some docs. I've tried to strike a balance: I want to tell people about this option, but not directly provide the code to avoid copy-paste without thinking.


camilonova:
@jacobian thank you. Glad to see you around here :)

Issue #376, opened 2020-02-10
Cause migrate bug with multi db


leench:
### Describe the problem

I have multi db with my project, when I migrate with non-default db, it raised::
`django.db.utils.ProgrammingError: relation "django_content_type" does not exist`
or
`django.db.utils.OperationalError: no such table: django_content_type`

### Steps to reproduce

Just use multi db and custom db router. 
full output:

```
  % ./manage.py migrate --database=test
Operations to perform:
  Apply all migrations: admin, auth, contenttypes, database, sessions, testf
Running migrations:
  No migrations to apply.
Traceback (most recent call last):
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/db/backends/sqlite3/base.py", line 383, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.OperationalError: no such table: django_content_type

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./manage.py", line 21, in <module>
    main()
  File "./manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/base.py", line 83, in wrapped
    res = handle_func(*args, **kwargs)
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/commands/migrate.py", line 257, in handle
    self.verbosity, self.interactive, connection.alias, apps=post_migrate_apps, plan=plan,
  File "/home/leen/tmp/env/lib/python3.7/site-packages/django/core/management/sql.py", line 51, in emit_post_migrate_signal
    **kwargs
[...]

PR #462, opened 2021-11-25
Add documentation for constance_dbs config

MansurAliKoroglu:
Documentation was missing for constance_dbs config.

If you have multiple databases and one of the DBs you have does not apply migrations for `contenttypes` then it is failing during migrating constance. Can be fixed with settings this option. It can be only found in changelog but there was no documentation.

Check #376 for example to the problem that I mentioned.


camilonova:
Thank you

PR #466, opened 2021-12-31
Add caching redis backend

horpto:
Quite often configuration is a rarely updated. And get current value in time is not so important. But redis backend always retrieves a current value.
Due to this simple access to the config has a sudden overhead especially in cycles. Programmer should always remeber about it and cache values manually.
Even a small timeout can drastically increase performance.

### DIFF
--- a/constance/backends/redisd.py
+++ b/constance/backends/redisd.py
@@ -1,5 +1,5 @@
 from pickle import loads, dumps
-from threading import Lock
+from threading import RLock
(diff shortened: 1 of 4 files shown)


camilonova:
Looks good to me. I only wonder what would happen in a multiple-deployment scenario. The value would be cached locally and not reflected on the other servers. Anyways, it will be up to the developer to take this into account.

@horpto can you check the failing tests?

Issue #454, opened 2021-10-07
Missing migration 0003_alter_constance_id


bt:
### Describe the problem

Whenever I run `migrate`, I receive the following message:

```
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

When I run the `makemigrations`, it tries to create a migration to `Alter field id on constance`. Is this normal?

### System configuration

* Django version: 3.2.8
* Python version: 3.9.7
* Django-Constance version: 2.8.0

bt:
Just as an addition, when I run it, the generated file contains the following operation:

```
    operations = [
        migrations.AlterField(
            model_name='constance',
            name='id',
            field=models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID'),
        ),
    ]
```

bt:
Oh, it might actually be due to the fact I set `DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'` in my settings, hopefully someone can confirm for me 😁

syurchenko:
Yes, it is because of it.

merwok:
I think this should stay open: constance should define default_auto_field = AutoField in its AppConfig to avoid depending on the project default.

Issue #452, opened 2021-09-15
ImportError: cannot import name 'Config' from partially initialized module 'constance.base' (most likely due to a circular import)


DrMeers:
As per https://stackoverflow.com/a/69185962/284164 :

Once every few hundred thousand requests I see one of these:

```
ImportError at /
cannot import name 'Config' from partially initialized module 'constance.base' (most likely due to a circular import) (/usr/local/lib/python3.9/site-packages/constance/base.py)
```

I cannot identify any rhyme or reason. It doesn't correspond with accessing `constance` admin, just randomly occurs. My best guess is it's something to do with the `LazyObject` in [`constance's __init__.py`](https://github.com/jazzband/django-constance/blob/master/constance/__init__.py), and maybe random race-conditions in restarting expired `gunicorn` workers or something?!

I'm using:
- Python 3.9.2
- Django 3.2
- `django-constance = {extras = ["database"],version = "==2.8.*"}`
- `"constance"` and `"constance.backends.database"` in `INSTALLED_APPS` (at top)
- `CONSTANCE_BACKEND = "constance.backends.database.DatabaseBackend"`
- `"constance.context_processors.config"` in `TEMPLATES[0]["OPTIONS"]["context_processors"]`

All my code does is `from constance import config` and access `config` attributes in the standard way in python code and Django templates.

For what it's worth, we've been using `django-constance` on this site for years, but never saw this error until we upgraded to `2.8.0` (from `2.6.0`). We were using Django 3.1 when it first appeared, but has also occurred since upgrading to 3.2.

----

I've managed to reproduce the issue and track down the likely cause.

I managed to trigger it once using Django's `runserver`, on the first request handled, but in hundreds of subsequent restarts/attempts I couldn't make it recur.

I then instead ran a couple of local `gunicorn` worker threads with a low requests-per-worker threshold, and spammed the local port with rapid-fire requests, and sure enough the `ImportError` occurs every now and then.

The issue seems to be this:

- `constance.__init__` uses a `django.utils.functional.LazyObject` which, once lazily evaluated, imports `constance.base` and instantiates a `Config` from that module.

- `constance.base.Config.__init__` uses `constance.utils.import_module_attr` to import the backend specified in the project settings, which in turn uses `importlib.import_module` to import, in my case, `constance.backends.database.DatabaseBackend`.

- `constance.backends.database.__init__` also imports `config `from `...` (`constance`), thus potentially creating an import loop.

[...]

Issue #414, opened 2020-07-05
Migration problem on multiple databases


euthygenes:
### Describe the problem
Constance tries to migrate my all databases and fails except `default` database.  It should work only `default` database. 

== fail message ==
`
django.db.utils.ProgrammingError: relation "django_content_type" does not exist
LINE 1: ..."."app_label", "django_content_type"."model" FROM "django_co..`

### Steps to reproduce
you can add a second database for some special models than try to migrate.

### System configuration

* Django version:
2.2.13
* Python version:
3.6
* Django-Constance version
2.7.0

WojtekNosal:
Use `CONSTANCE_DBS` in settings to limit to certain DBs
[link to changelong](https://django-constance.readthedocs.io/en/latest/changes.html?highlight=CONSTANCE_DBS#v2-2-0-2018-03-23)

PR #469, opened 2022-01-11
Refactor app and model

sergei-iurchenko:
This MR reconsiders model status in the constance application. 
In current state model _Constance_ does not belong to application _constance_ - it belongs to application _database_.
This have several issues:
1) Impossible to support content type and permission management via Django build-in functions.
ConstanceConfig contains code to create them manually. Django does it automatically.
Build-in cleaner of stale content types _./manage.py remove_stale_contenttypes --include-stale-apps_ removes manually created permissions and content types

2) https://github.com/jazzband/django-constance/issues/454
My PR fixes it
3) Different import errors which some users get. 
https://github.com/jazzband/django-constance/issues/452
Such issues would go away.
4) Enables Django build-in database routing to work by default
https://github.com/jazzband/django-constance/issues/414 
https://github.com/jazzband/django-constance/issues/376

What changes?
Database table would always be created. It would be empty if DatabaseBackend is not used.
Migration on new table would be automatic for new and old installations.

Django is carp enough for model to be attached to application. django-constance is rather old application and earlier it was not so strict.

This PR is draft. It is working example with code refactoring tested on 2 projects and postgres database.
If my refactoring can be accepted I would add documentation, release notes about migration and will test on sqlite, mysql and postgres carefully.

I am open to discuss MR


tox.ini
```diff
@@ -8,7 +8,6 @@ envlist =
 deps =
     redis
     coverage
-    mock
```
syurchenko, line 11: mock for python 2. in python 3 it is build in

tests/test_app.py
```diff
@@ -1,56 +0,0 @@
-from django.apps import apps
```
syurchenko, line 1: this file is useless now. django does all staff automatically

syurchenko:
Hello @Mogost !
Review please =)

syurchenko:
Hello @Mogost !
Sorry for so long review fixes. Now everything is saved: table, permissions, content types.

Mogost:
@jazzband/django-constance Please help to review

coverage.xml
```diff
@@ -0,0 +1,712 @@
+<?xml version="1.0" ?>
```
camilonova, line 1: This file should be ignored right?

docs/index.rst
```diff
@@ -37,6 +37,7 @@ the :setting:`CONSTANCE_CONFIG` section, like this:
         'django.contrib.messages',
         ...
         'constance',
+        'constance.backends.database',
```
camilonova, line 40: This will make it harder to use/understand. What happened with the backend setting?

camilonova (commented):
I strongly don't like the extra app in the INSTALLED_APPS.

docs/index.rst
syurchenko: This application was added to INSTALLED_APPS for migrations consistency. 
If I remove it then migration to new app would not be transparent for users of this battery(and it can break there own migrations as Mogost mentioned)
The way how it is made now version updating is transparent.

Nothing has happened with backend setting. It works as usual.

Do you have any thought how to make it better?

camilonova: @Mogost how do you think we can get the same result without adding this extra line?

Mogost: I am not sure. May be `replaces` option in `constance` app will help.

syurchenko: What is **replaces** option?

Mogost: It's what django uses on squashmigrations https://docs.djangoproject.com/en/4.0/topics/migrations/#squashing-migrations
After squashing you will have replace the attribute in migration.

Mogost:
```
    replaces = [
        ("migrations", "0001_initial"),
        ("migrations", "0002_second"),
    ]

```

syurchenko: I have checked this variant. Squashing of migrations is only possible within one application [only possible within one application ](https://docs.djangoproject.com/en/4.0/ref/django-admin/#django-admin-squashmigrations).
In our case there are 2 applications: database & constance.

syurchenko: @Mogost any thoughts?

Mogost: Nothing from my side.
I don't know how to do it better.
May be we can forget about my point [here](https://github.com/jazzband/django-constance/pull/469?notification_referrer_id=NT_kwDOAAsEK7EyOTMzNDU4MjQ4OjcyMTk2Mw#issuecomment-1040508408)
> Just for example, I found a place in my project where my migrations were dependent on Constance migrations like that:
> ```python
> class Migration(migrations.Migration):
>     dependencies = [
>        ('database', '0001_initial'),
>    .....
>```

We can forget about this and bump the major version with a few words in release notes.

Mogost: @camilonova  It seems that we already have `constance.backends.database` in our [docs here](https://django-constance.readthedocs.io/en/latest/backends.html#database). So it turns out it's always been there.

This was probably done with the assumption that migration is unnecessary if the data is stored in redis and not in the database.

@jezdez is the original author of this code. And we probably need his advice here. I have just mailed him.
We are now actually changing the functionality that could lead to data loss for our users. We need to be careful with this pr.

camilonova: I agree. At this point, I'm not in favor of making people add two lines at the INSTALLED_APPS setting to use this library.

syurchenko: Finally I merge develop, fix conflicts and you will be able to merge MR?

sergei-iurchenko:
@Mogost @camilonova Well, I fixed everything. Review finally please =)

camilonova (approved):

constance/apps.py
```diff
@@ -1,45 +1,9 @@
-from django.db.models import signals
-from django.apps import apps, AppConfig
+from django.apps import AppConfig
 from django.utils.translation import gettext_lazy as _
 
 
 class ConstanceConfig(AppConfig):
     name = 'constance'
     verbose_name = _('Constance')
+    default_auto_field = 'django.db.models.AutoField'
```
zodman, line 8:
```suggestion
    default_auto_field =getattr(settings, "DEFAULT_AUTO_FIELD", 'django.db.models.AutoField')
```

why not this could be overrrited by a variables from setting.

sergei-iurchenko: Do you mean default django settings or project settings? In both cases things can change and generate extra migration in installed dependencies what is not  expected

SpecialK118:
Does this mean there's no meow database backend for constance?
I upgraded from 2.9.1 and I'm unable to get it to work.
When I try to install `django-constance[database]` I get:
`WARNING: django-constance 3.1.0 does not provide the extra 'database'`

When I try to run, I get:
`    from constance.backends.database.models import Constance
ModuleNotFoundError: No module named 'constance.backends.database.models'; 'constance.backends.database' is not a package`

sergei-iurchenko:
@SpecialK118 Hello) install just via `pip install django-constance`.
'`constance.backends.database`' should be removed from INSTALLED_APPS
Changes are described in [changelog](https://github.com/jazzband/django-constance/blob/master/docs/changes.rst)

PR #475, opened 2022-03-17
Adding support for using a subdirectory of `MEDIA_ROOT` for file fields

### LINKED ISSUE #334: How can i add a subfolder for file uploads?
I want to set a subfolder after my MEDIA_ROOT for each image_field item. Otherwise it's saving all files to MEDIA_ROOT and it doesn't satisfy my requirements.

### Steps to reproduce
We can add a special parameter like upload_to and this parameter is helps us for setting subfolder. If anyone directs me i can contribute for it.

### System configuration

* Django version: 1.11.22
* Python version: 2.7.16

### LINKED ISSUE #398: Constance upload path
### Describe the problem

I want to set upload path to constance images and files. How can I do it?

### System configuration

* Django version: 2.2
* Python version: 3.6

BigglesZX:
A couple of [historic](https://github.com/jazzband/django-constance/issues/334) [issues](https://github.com/jazzband/django-constance/issues/398) refer to the behaviour of `FileField` or `ImageField` instances used as part of `CONSTANCE_ADDITIONAL_FIELDS`, which will upload their contents directly to the root of `MEDIA_ROOT`.

I recently ran into this myself and wanted to be able to organise those files in a subdirectory as I do with other media in the project uploaded via model fields (e.g. using `upload_to`).

The library uses [`default_storage.save()`](https://docs.djangoproject.com/en/4.0/topics/files/#storage-objects) to save uploaded files – I did some testing to ensure that if a relative path was prepended to the uploaded file name, it would be accepted – and it is, as the base `Storage` class will accept relative paths here. I can't verify that this is true of every storage engine.

Therefore this PR adds a simple `os.path.join` to add a new setting, `CONSTANCE_FILE_ROOT`, to the front of any uploaded filename, which has the effect of saving the file in that directory inside `MEDIA_ROOT`. This path is what is persisted to the database so any lookups of the field will include the setting. Therefore existing retrieval behaviour (see the example in the docs that uses a field called `constance.LOGO_IMAGE`) is preserved. If the setting is empty, `join()` will return the original filename.

I've added a test image field to the example project in the repository and run local tests. I haven't contributed to this library before so let me know if there's more I can do. Thanks!

Resolves #334, resolves #398.


camilonova (approved):

BigglesZX:
Any chance of getting this merged in?

camilonova:
@BigglesZX can you fix the conflicts?

BigglesZX:
@camilonova Thanks, done!

camilonova (approved):

PR #499, opened 2022-09-21
Forward the request when saving the form

felippem:
Hello. I would like to forward the request to the instance of ConstanceForm, when calling the save method. Help me understand if it is necessary to update any tests.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,7 +1,12 @@
 Changelog
 ---------
 
-2.9.1 (2022-08-11)
+v2.10.0 (unreleased)
+~~~~~~~~~~~~~~~~~~
+
+* Forward the request when saving the admin changelist form
+
+v2.9.1 (2022/08/11)
 ~~~~~~~~~~~~~~~~~~
 
 * Add support for gettext in fieldset headers
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -250,7 +250,7 @@ class ConstanceAdmin(admin.ModelAdmin):
         form = form_cls(initial=initial, request=request)
         if request.method == 'POST' and request.user.has_perm('constance.change_config'):
             form = form_cls(
-                data=request.POST, files=request.FILES, initial=initial
+                data=request.POST, files=request.FILES, initial=initial, request=request
             )
             if form.is_valid():
                 form.save()
--- a/tests/test_admin.py
+++ b/tests/test_admin.py
@@ -8,11 +8,11 @@ from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseRedirect
 from django.template.defaultfilters import linebreaksbr
 from django.test import TestCase, RequestFactory
+from django.utils.translation import gettext_lazy as _
 
 from constance import settings
-from constance.admin import Config
-from constance.admin import get_values
-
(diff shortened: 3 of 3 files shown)


felippem:
@Mogost @jezdez @Mogost Can you please review this?

jezdez:
@felippem Yes, the tests need to be updated, but I'm also curious for the use case?

felippem:
> @felippem Yes, the tests need to be updated, but I'm also curious for the use case?

@jezdez I'd like to get the request.user.has_perm when the form loads, after it's posted.  I have a custom form, which inherits Constance's original implementation.

felippem:
> > @felippem Yes, the tests need to be updated, but I'm also curious for the use case?
> 
> @jezdez I'd like to get the request.user.has_perm when the form loads, after it's posted.  I have a custom form, which inherits Constance's original implementation.

Hi @jezdez ! Did you understand my use case?

Mogost (approved):

jezdez:
@felippem I get it now, but this is actually a regression, since there is a [permission check in the `ConstanceForm.__init__` method](https://github.com/jazzband/django-constance/blob/02c5fd5011d8557f99af196bc9f73fa1df001eb4/constance/admin.py#L105) that is currently not run since the request isn't passed. Since the tests don't fail, it seems as if this line of code isn't accurately tested. Please add/update a test case for this.

jezdez:
This was an oversight in https://github.com/jazzband/django-constance/pull/393.

felippem:
> @felippem I get it now, but this is actually a regression, since there is a [permission check in the `ConstanceForm.__init__` method](https://github.com/jazzband/django-constance/blob/02c5fd5011d8557f99af196bc9f73fa1df001eb4/constance/admin.py#L105) that is currently not run for POST requests since the request isn't passed. Since the tests don't fail, it seems as if this line of code isn't accurately tested. Please add/update a test case for this.

So I'll try to do that soon.  Thanks for answering.

felippem:
> > @felippem I get it now, but this is actually a regression, since there is a [permission check in the `ConstanceForm.__init__` method](https://github.com/jazzband/django-constance/blob/02c5fd5011d8557f99af196bc9f73fa1df001eb4/constance/admin.py#L105) that is currently not run for POST requests since the request isn't passed. Since the tests don't fail, it seems as if this line of code isn't accurately tested. Please add/update a test case for this.
> 
> So I'll try to do that soon. Thanks for answering.

@jezdez Done. Can you review please?

docs/changes.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+2.10.0 (2022-10-08)
+~~~~~~~~~~~~~~~~~~
+
+* Regression fix (#393) - forward the request when saving the form
+
```
felippem, line 8: @jezdez What do you think about this?

felippem:
Thank you @jezdez. Have an idea about the next release?

felippem:
@jezdez can I help with anything else?

camilonova (approved):

camilonova:
Thank you

felippem:
Hi, @camilonova! Do you have a date for the next release?

PR #512, opened 2023-03-10
fix_ci

sergei-iurchenko:
CI was failing for last months because new version of **tox** (4.0) appeared.
I could not support this version but I fixed old version, deprecated old versions of python, django.
May be somebody will be able to move project to **tox** 4.0 but it is not target of this PR.


camilonova (approved):

camilonova:
Thank you

Issue #510, opened 2023-02-04
Data loss using DatabaseBackend when db connection unstable. 


a-edakin:
### Describe the problem

I noticed that once a month or two some most used constance would just reset to default values. 

After some insvestigation I found out that it was caused when db connection was unstable (db was overloaded or restarted for example)

### Steps to reproduce

I don't think it would be easy to reproduce because of percise timings but it can be reproduced with some mocks in tests.

In method get of DatabaseBackend https://github.com/jazzband/django-constance/blob/master/constance/backends/database/__init__.py#L69
```
class DatabaseBackend(Backend):
# ...

    def get(self, key):
        key = self.add_prefix(key)
        if self._cache:
            value = self._cache.get(key)
            if value is None:
                self.autofill()
                value = self._cache.get(key)
        else:
            value = None
        if value is None:
            try:
                value = self._model._default_manager.get(key=key).value
            except (OperationalError, ProgrammingError, self._model.DoesNotExist):
                pass
            else:
                if self._cache:
                    self._cache.add(key, value)
        return value
```

As you can see when we try to `get` value from database we just ignore exceptions OperationalError and ProgrammingError. Let's assume db was overloaded and caused OperationalError for example so when we return value it is None.

Now we go to wrapper class Config https://github.com/jazzband/django-constance/blob/master/constance/base.py#L20

```
class Config:
# ...
    def __getattr__(self, key):
        try:
            if not len(settings.CONFIG[key]) in (2, 3):
                raise AttributeError(key)
            default = settings.CONFIG[key][0]
        except KeyError:
            raise AttributeError(key)
        result = self._backend.get(key)
        if result is None:
            result = default
            setattr(self, key, default)
            return result
        return 
```

Just like I said before method `get` returns None and in this case wrapper `Config.__getattr__` just sets value too default and at this time all database issues are resolved and that is how we just lost data of a constance.

How I fixed issue right now:

I just inherited DatabaseBackend and removed OperationalError and ProgrammingError from `except` block
```
class CustomDatabaseBackend(DatabaseBackend):

    def get(self, key):
        key = self.add_prefix(key)
        if self._cache:
            value = self._cache.get(key)
            if value is None:
                self.autofill()
                value = self._cache.get(key)
        else:
            value = None
        if value is None:
            try:
                value = self._model._default_manager.get(key=key).value
            except self._model.DoesNotExist:
                pass
            else:
                if self._cache:
                    self._cache.add(key, value)
        return value
```

### System configuration

[...]

PR #525, opened 2023-07-21
allow the OperationalError to be raised

rudolfolah:
when trying to get a value while there is no database connection, the OperationalError will be raised. We should allow it to be raised instead of suppressing it.

Related to https://github.com/jazzband/django-constance/issues/510

### DIFF
--- a/constance/backends/database.py
+++ b/constance/backends/database.py
@@ -78,7 +78,7 @@ class DatabaseBackend(Backend):
         if value is None:
             try:
                 value = self._model._default_manager.get(key=key).value
-            except (OperationalError, ProgrammingError, self._model.DoesNotExist):
+            except (ProgrammingError, self._model.DoesNotExist):
                 pass
             else:
                 if self._cache:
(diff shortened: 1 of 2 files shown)


DrMeers:
the `OperationalError` handling was added in https://github.com/whs/django-constance/commit/bbfa8fa7dc28d67c8a8721b543c7c41bc99785c4 for https://github.com/jazzband/django-constance/issues/229 so we need to ensure this won't create regressions there

sergei-iurchenko:
Actually https://github.com/jazzband/django-constance/issues/229 is rather strange PR. There were some troubles with migrations in that project.
And also I have refactored constance application for modern django way so everything should work fine

sergei-iurchenko:
@camilonova I suggest to skip tests and merge only changes in code. Tests would not show any regression. This exception should not be silenced.

camilonova:
I'm going to close this one because of the merge conflicts. If a related issue comes up this could be used as a reference for a new PR.

Issue #528, opened 2023-08-22
data loss upgrading from 2.9.1 to 3.1.0 with database backend


mfisco:
### Describe the problem

Upgrading from 2.9.1 to 3.1.0 and running `python manage migrate`, both migrations appear successful, but the data isn't transferred, nor is the old table dropped.

### Steps to reproduce

1. Install version 2.9.1 and make a couple of sample settings, changing from their defaults.
2. Update to 3.1.0 and run `python manage.py migrate`
3. Inspect the DB table and notice the new `constance_constance` table created from `0001_initial` is empty, and the old `constance_config` table still exists.

After some debugging, I think it concerns the SQL in `0002_migrate_from_old_table`. Specifically, If I remove the try/except block, I get the following exception:

```
django.db.utils.ProgrammingError: (1064, "You have an error in your SQL syntax; 
check the manual that corresponds to your MySQL server version for the right syntax
to use near 'key, value ) SELECT id, key, value FROM constance_config' at line 1")
```

But by escaping the column names (e.g. ``` ( id, key, value )``` → ```( `id`,  `key`, `value` )```), I was able to get the 2nd migration to work (data transferred/old table dropped), leading me to believe it's because of SQL reserved words. 

### System configuration

* Django version: 4.2.4
* Python version: 3.10.6
* Django-Constance version: 3.1.0
* MySQL version: 8.0.18

sergei-iurchenko:
It will be fixed soon https://github.com/jazzband/django-constance/pull/531
Sorry, my fault

mfisco:
No worries and thanks for your help!

Volody2006:
It seems that fixing this error caused an error for me.

System configuration
Django version: 3.1
Python version: 3.9
Django-Constance version: 3.1.0
MySQL version: 8.0.22

Using `connection = schema_editor.connection` does not work in my case.

jasisz:
I have just experienced the very same issue with MySQL, fixed it on my own and later found this issue...

If anyone lands in here looking for the answer "why": this migration may be indeed fixed, but it is not yet released.

I am not sure if the fix works for me either, I had to run this sqls manually to fix it (state was already messed up a bit):
```
DELETE FROM constance_constance;
INSERT INTO constance_constance SELECT * FROM constance_config;
DROP TABLE constance_config;
```

mfisco:
On that note, would it be possible to generate a new release with the corrected migration file?

stitch:
[...]

PR #531, opened 2023-10-05
fix migration on mysql

sergei-iurchenko:
https://github.com/jazzband/django-constance/issues/528
Quoted is done correctly now.
I have tested on sqlite, mysql & postgresql

### DIFF
--- a/constance/migrations/0002_migrate_from_old_table.py
+++ b/constance/migrations/0002_migrate_from_old_table.py
@@ -1,5 +1,10 @@
+from logging import getLogger
+
 from django.core.management.color import no_style
-from django.db import migrations, connection, DatabaseError
+from django.db import migrations, DatabaseError
+
+
+logger = getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


sergei-iurchenko:
@camilonova review please

camilonova (approved):

jasisz:
I have just experienced the very same issue with MySQL, fixed it on my own and later found this issue...
If anyone lands in here looking for the answer "why": this migration may be indeed fixed, but it is not yet released. I am not sure if the fix works for me either, I had to run this sqls manually to fix it (state was already messed up a bit):
```
DELETE FROM constance_constance;
INSERT INTO constance_constance SELECT * FROM constance_config;
DROP TABLE constance_config;
```

schefDev:
First of all, thank you very much for your great work of the entire jazzband community.  💖

Just a itty-bitty note, this fix is of course essential as there is a risk of data loss, but unfortunately it has not yet been released.

PR #538, opened 2024-02-01
Fix issue #510

### LINKED ISSUE #510: Data loss using DatabaseBackend when db connection unstable.

chrisclark:
Fixes #510

### DIFF
--- a/constance/backends/database.py
+++ b/constance/backends/database.py
@@ -76,11 +76,11 @@ class DatabaseBackend(Backend):
         else:
             value = None
         if value is None:
-            try:
-                value = self._model._default_manager.get(key=key).value
-            except (OperationalError, ProgrammingError, self._model.DoesNotExist):
+            match = self._model._default_manager.filter(key=key)
+            if len(match) == 0:
                 pass
             else:
+                value = match.first().value
                 if self._cache:
                     self._cache.add(key, value)
         return value
--- a/constance/backends/database.py
+++ b/constance/backends/database.py
@@ -68,19 +68,16 @@ class DatabaseBackend(Backend):
 
     def get(self, key):
         key = self.add_prefix(key)
+        value = None
         if self._cache:
             value = self._cache.get(key)
             if value is None:
                 self.autofill()
                 value = self._cache.get(key)
-        else:
-            value = None
         if value is None:
-            match = self._model._default_manager.filter(key=key)
-            if len(match) == 0:
-                pass
-            else:
-                value = match.first().value
+            match = self._model._default_manager.filter(key=key).first()
+            if match:
+                value = match.value
                 if self._cache:
                     self._cache.add(key, value)
         return value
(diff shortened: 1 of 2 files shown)


lindsayemarc (approved):
looks good to me. I am in favor of this approach compared to simply removing the operationalerror from the except statement. removing the try catch entirely and letting all the errors raise side steps the issue of setting the value when an error is raised.

Anton-Shutik:
@jezdez Hi, are there any plans to release it to pypi ?

PR #544, opened 2024-06-08
chore: migrate to f-string

Rotzbua:
Use current style to print variables in strings.

credit: tool pyupgrade

### DIFF
--- a/constance/management/commands/constance.py
+++ b/constance/management/commands/constance.py
@@ -60,7 +60,7 @@ class Command(BaseCommand):
 
         if command == 'get':
             try:
-                self.stdout.write("{}".format(getattr(config, key)), ending="\n")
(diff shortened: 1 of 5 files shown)


constance/management/commands/constance.py
```diff
@@ -60,7 +60,7 @@ def handle(self, command, key=None, value=None, *args, **options):
 
         if command == 'get':
             try:
-                self.stdout.write("{}".format(getattr(config, key)), ending="\n")
+                self.stdout.write(f"{getattr(config, key)}", ending="\n")
```
Mogost, line 63:
```suggestion
                self.stdout.write(str(getattr(config, key)), ending="\n")
```

Mogost (approved):
Thank you. Looks good!

PR #545, opened 2024-06-08
chore: update supported versions

### LINKED ISSUE #540: Support for Django 5
### Describe the problem

The latest Django version (v5+) appears to be unsupported.

### Steps to reproduce

Django 5 has been released, but it does not appear to be supported according to the README.

Rotzbua:
## Changes

- drop eol django 3.2 4.0 4.1
- drop eol python 3.7
- drop untested PyPy support
- prepare python 3.12 test
- prepare django 5.1 test
- update GH action dependencies
- add dependabot to maintain GH action dependencies

## Reference

fixes #540

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore(ci): "
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    open-pull-requests-limit: 1
--- a/tox.ini
+++ b/tox.ini
@@ -2,20 +2,18 @@
 requires =
     tox<4
 envlist =
-    py{37,38,39,310}-dj{32}-{unittest,pytest}
-    py{38,39,310}-dj{40}-{unittest,pytest}
-    py{38,39,310,311}-dj{41,42}-{unittest,pytest}
-    py{310,311}-dj{main}-{unittest,pytest}
+    py{38,39,310,311,312}-dj{42}-{unittest,pytest}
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore(ci): "
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    open-pull-requests-limit: 1
```
Mogost, line 18:
```suggestion
```

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    commit-message:
+      prefix: "chore(ci): "
+    groups:
+      github-actions:
+        patterns:
+          - "*"
```
Mogost, line 17: Isn't it better to do this without grouping?
```suggestion
```

tox.ini
```diff
@@ -34,8 +33,8 @@ setenv =
 
 [gh-actions]
 python =
-    3.7: py37
     3.8: py38
     3.9: py39
     3.10: py310
     3.11: py311
+    3.12: py312
```
Mogost, line 40: How does this correspond with?
```
.github/workflows/test.yml

python-version: ['3.8', '3.9', '3.10', '3.11'] # '3.12' does not work with tox@3
```

tox.ini
```diff
@@ -2,20 +2,19 @@
 requires =
     tox<4
 envlist =
-    py{37,38,39,310}-dj{32}-{unittest,pytest}
-    py{38,39,310}-dj{40}-{unittest,pytest}
-    py{38,39,310,311}-dj{41,42}-{unittest,pytest}
-    py{310,311}-dj{main}-{unittest,pytest}
+    py{38,39,310,311,312}-dj{42}-{unittest,pytest}
+    py{310,311,312}-dj{50}-{unittest,pytest}
+    #py{310,311,312}-dj{51}-{unittest,pytest}
```
Mogost, line 7: Is it a coincidence that you commented on that line?

Mogost: I don't think we should add the commented configuration for the future. it is more logical to create a Pull request after the release of this version, which will contain the necessary set of changes.

Mogost (approved):

PR #559, opened 2024-07-02
Improve styling, add css classes to admin templates

faebebin:
makes this ...
image

look like this ...
image

### DIFF
--- a/constance/templates/admin/constance/change_list.html
+++ b/constance/templates/admin/constance/change_list.html
@@ -48,7 +48,7 @@
 
             {% if fieldsets %}
               {% for fieldset in fieldsets %}
-              <fieldset class="module{% if fieldset.collapse %} collapse{% endif %}">
+              <fieldset class="module{% if fieldset.collapse %} collapse{% endif %} card card-body">
                   <h2>{{ fieldset.title }}</h2>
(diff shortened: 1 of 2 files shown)


constance/templates/admin/constance/change_list.html
```diff
@@ -48,7 +48,7 @@
 
             {% if fieldsets %}
               {% for fieldset in fieldsets %}
-              <fieldset class="module{% if fieldset.collapse %} collapse{% endif %}">
+              <fieldset class="module{% if fieldset.collapse %} collapse{% endif %} card card-body">
```
Mogost, line 51: This is not from standard django.

Mogost (changes requested):

Mogost:
You shouldn't use styles that are not available in base Django https://github.com/django/django/tree/main/django/contrib/admin/static/admin/css

constance/templates/admin/constance/change_list.html
faebebin: ah ok, yes we have bootstrap in there I guess. Feel free to close this pull request and open an issue if the bad styling actually appears also on other instances.

will come back to it hopefully later but not have time to work on it right now, sry

PR #560, opened 2024-07-03
Add ruff format & lint (isort only)

Mogost:
1. Applied formatting
2. Configured ruff linter but enabled only isort rules to reduce changes (more linters will be enabled separately)
3. Added linters in ci

_All code changes were made by ruff automatically and marked by it as safe_

### DIFF
--- a/constance/backends/database.py
+++ b/constance/backends/database.py
@@ -1,21 +1,22 @@
 from django.core.cache import caches
 from django.core.cache.backends.locmem import LocMemCache
 from django.core.exceptions import ImproperlyConfigured
-from django.db import (
-    IntegrityError,
-    OperationalError,
-    ProgrammingError,
-    transaction,
-)
+from django.db import IntegrityError
+from django.db import OperationalError
+from django.db import ProgrammingError
+from django.db import transaction
 from django.db.models.signals import post_save
 
+from constance import config
+from constance import settings
+from constance import signals
 from constance.backends import Backend
-from constance import settings, signals, config
 
 
 class DatabaseBackend(Backend):
     def __init__(self):
         from constance.models import Constance
+
         self._model = Constance
         self._prefix = settings.DATABASE_PREFIX
         self._autofill_timeout = settings.DATABASE_CACHE_AUTOFILL_TIMEOUT
@@ -24,16 +25,17 @@ class DatabaseBackend(Backend):
         if self._model._meta.app_config is None:
             raise ImproperlyConfigured(
                 "The constance.backends.database app isn't installed "
-                "correctly. Make sure it's in your INSTALLED_APPS setting.")
+                "correctly. Make sure it's in your INSTALLED_APPS setting."
+            )
 
         if settings.DATABASE_CACHE_BACKEND:
             self._cache = caches[settings.DATABASE_CACHE_BACKEND]
             if isinstance(self._cache, LocMemCache):
                 raise ImproperlyConfigured(
-                    "The CONSTANCE_DATABASE_CACHE_BACKEND setting refers to a "
+                    'The CONSTANCE_DATABASE_CACHE_BACKEND setting refers to a '
                     "subclass of Django's local-memory backend (%r). Please "
-                    "set it to a backend that supports cross-process caching."
-                    % settings.DATABASE_CACHE_BACKEND)
+                    'set it to a backend that supports cross-process caching.' % settings.DATABASE_CACHE_BACKEND
+                )
         else:
             self._cache = None
         self.autofill()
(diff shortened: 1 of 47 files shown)


Natim (approved):

pyproject.toml
```diff
@@ -54,3 +54,23 @@ changelog = "https://github.com/jazzband/django-constance/releases/"
 
 [tool.setuptools.packages.find]
 include = ["constance*"]
+
+
+[tool.ruff]
+# Same as Black.
+line-length = 88
```
Natim, line 61:
```suggestion
line-length = 99
```

pyproject.toml
```diff
@@ -54,3 +54,23 @@ changelog = "https://github.com/jazzband/django-constance/releases/"
...
+indent-width = 4
+
+[tool.ruff.format]
+quote-style = "double"
```
sergei-iurchenko, line 65: project is mostly written with single quotes, let`s leave status quo for this.

pyproject.toml
```diff
@@ -54,3 +54,23 @@ changelog = "https://github.com/jazzband/django-constance/releases/"
...
+skip-magic-trailing-comma = false
+line-ending = "auto"
+
+[tool.ruff.lint]
+select = [
+    "I", # isort
+]
+
+[tool.ruff.lint.isort]
+force-single-line = true
```
sergei-iurchenko, line 76: What is the reason to do so?

For me
**from datetime import date, datetime, time, timedelta**
is better then 
**from datetime import date
from datetime import datetime
from datetime import time
from datetime import timedelta**

Natim: It's the trend nowadays because it's easy to remove a line when the import is removed.

pyproject.toml
Natim: Most Python project I know use double quote and tabs, doesn't it?

pyproject.toml
Mogost: The single-line format reduces merge conflicts significantly

pyproject.toml
Mogost: In black it's 88. Also, 120 is commonly used. 99 is something not usual. I prefer to leave 88 or change to 120

Natim: AFAIK the PEP8 never talks about 88 or 120, I was suggesting 99 because that's the maximum the PEP8 talks about. I'm not against 120 either but it seems pretty uncommon to me, should it be 119 to account for the `\n`?

pyproject.toml
Mogost: It doesn't matter to me https://github.com/jazzband/django-constance/pull/560/commits/a40c0ab7c2f14fe19744a59d9ec0f81a85596fce
changed

Mogost: can revert again if there is sense 😄

pyproject.toml
Natim: The goal is mostly to reduce this kind of changes:

Mogost: @Natim Can we agree on 120?

Mogost: https://github.com/jazzband/django-constance/pull/560/commits/20c9d7765e5c691e744dabccd6161262bcc54634 lets try

Mogost: Just applied formating from scratch to avoid some changes. _forcepushed_

Mogost: Also about the code on your screen. I already have a refactoring of this command in my stash.

constance/backends/database.py
```diff
@@ -24,24 +25,25 @@ def __init__(self):
...
             self._cache = caches[settings.DATABASE_CACHE_BACKEND]
             if isinstance(self._cache, LocMemCache):
                 raise ImproperlyConfigured(
-                    "The CONSTANCE_DATABASE_CACHE_BACKEND setting refers to a "
+                    'The CONSTANCE_DATABASE_CACHE_BACKEND setting refers to a '
                     "subclass of Django's local-memory backend (%r). Please "
-                    "set it to a backend that supports cross-process caching."
-                    % settings.DATABASE_CACHE_BACKEND)
+                    'set it to a backend that supports cross-process caching.' % settings.DATABASE_CACHE_BACKEND
+                )
```
Natim, line 38: Should we take the opportunity to migrate this to a f-string?

pyproject.toml
Natim: Yes it looks good to me like that indeed.

constance/backends/database.py
Mogost: We should. There are linters for that in ruff. But to reduce complexity of reviewing process I've moved this out of scope of this PR.

Natim: Looks like the start of a great refresh :pray:

Mogost: OK. than go next 😄

pyproject.toml
sergei-iurchenko: I feel the reason, thank you. I like this trend.

PR #562, opened 2024-07-03
Enable more rules for ruff

Mogost:


constance/admin.py
```diff
@@ -35,8 +35,8 @@ def __init__(self, model, admin_site):
     def get_urls(self):
         info = self.model._meta.app_label, self.model._meta.module_name
         return [
-            path('', self.admin_site.admin_view(self.changelist_view), name='%s_%s_changelist' % info),
-            path('', self.admin_site.admin_view(self.changelist_view), name='%s_%s_add' % info),
+            path('', self.admin_site.admin_view(self.changelist_view), name=f'{info[0]}_{info[1]}_changelist'),
+            path('', self.admin_site.admin_view(self.changelist_view), name=f'{info[0]}_{info[1]}_add'),
         ]
```
Natim, line 40:
```suggestion
        info = f"{self.model._meta.app_label}_{self.model._meta.module_name}"
        return [
            path('', self.admin_site.admin_view(self.changelist_view), name=f'{info}_changelist'),
            path('', self.admin_site.admin_view(self.changelist_view), name=f'{info}_add'),
        ]
```

constance/admin.py
```diff
@@ -125,17 +114,20 @@ def changelist_view(self, request, extra_context=None):
...
                     fields_list = fieldset_data
                     collapse = False
 
                 absent_fields = [field for field in fields_list if field not in settings.CONFIG]
-                assert not any(absent_fields), (
-                    'CONSTANCE_CONFIG_FIELDSETS contains field(s) that does ' 'not exist: %s' % ', '.join(absent_fields)
-                )
+                if any(absent_fields):
+                    raise ValueError(
+                        'CONSTANCE_CONFIG_FIELDSETS contains field(s) that does not exist(s): {}'.format(
```
Natim, line 127: :+1:

constance/apps.py
```diff
@@ -8,4 +8,4 @@ class ConstanceConfig(AppConfig):
     default_auto_field = 'django.db.models.AutoField'
 
     def ready(self):
-        from . import checks
+        pass
```
Natim, line 11:
```suggestion
```

Natim: To be honest, I believe we should keep that, it activates the checks.

constance/backends/redisd.py
```diff
@@ -35,7 +36,7 @@ def add_prefix(self, key):
     def get(self, key):
         value = self._rd.get(self.add_prefix(key))
         if value:
-            return loads(value)
+            return loads(value)  # noqa: S301
```
Natim, line 39: That's exactly why I don't like using Pickle for that...
1. It's not portable in between Python version
2. It's not portable in between languages, hard to read when you look at your redis, etc.

constance/test/__init__.py
```diff
@@ -1 +1,3 @@
 from .unittest import override_config  # pragma: no cover
+
+__all__ = ['override_config']
```
Natim, line 3: :+1:

tests/test_cli.py
```diff
@@ -18,10 +19,8 @@ def setUp(self):
         self.out = StringIO()
 
     def test_help(self):
-        try:
+        with contextlib.suppress(SystemExit):
```
Natim, line 22: :open_mouth: I'm impressed

Natim:
Good job :+1:

constance/backends/redisd.py
Mogost: Faced this in my project when we need read constance value in another service from db. But service is that written in another language. This should be rethinked

constance/apps.py
Mogost: Checks are activated because of registeing via `register`

```
from django.core.checks import Error, register
@register()
....
```

Natim: Yes but you need to call register and I believe the `ready()` call does load the code and trigger register.

constance/admin.py
Mogost: https://github.com/jazzband/django-constance/pull/562/commits/94cbb602d1e03e3eaab4420a04382985881d6050

constance/apps.py
Mogost: It's not so.
While you use register and place `"constance"` in `INSTALLED_APPS` you do not need to do more (like import)

Natim: Ok thank you for checking :pray:

Mogost: I have tested with my project
```
❯ ./manage.py check --tag constance
System check identified no issues (0 silenced).
❯ ./manage.py check --tag not_constance
CommandError: There is no system check with the "not_constance" tag.
```

Mogost: https://github.com/jazzband/django-constance/pull/562/commits/d30759be5d94122054f95827e720d0654dcb3267

constance/backends/redisd.py
Mogost: As an idea. 
I think it's possible to use modified `django.db.models.JSONField`.
And saving something like 
```
{
	"type": "int",
	"value": "1",
}
```
And casting it to the right value in to_reprenetaation

tests/backends/test_database.py
```diff
@@ -13,15 +13,15 @@ def setUp(self):
     def test_database_queries(self):
         # Read and set to default value
         with self.assertNumQueries(5):
-            self.config.INT_VALUE
+            _ = self.config.INT_VALUE
```
sergei-iurchenko, line 16: this artificial assigments look very noisy, let`t disable this lint. 
(and several more places with this)

constance/apps.py
```diff
@@ -6,6 +6,3 @@ class ConstanceConfig(AppConfig):
     name = 'constance'
     verbose_name = _('Constance')
     default_auto_field = 'django.db.models.AutoField'
-
-    def ready(self):
-        from . import checks
```
sergei-iurchenko, line 11: this should be returned to enable checks
```
    def ready(self):
        from . import checks  # noqa: F401
```

constance/backends/redisd.py
```diff
@@ -23,7 +24,7 @@ def __init__(self):
             try:
                 import redis
             except ImportError:
-                raise ImproperlyConfigured('The Redis backend requires redis-py to be installed.')
+                raise ImproperlyConfigured('The Redis backend requires redis-py to be installed.') from None
```
sergei-iurchenko, line 27: raises from None actually just silent linter. I suppose some linter bug. Correct usage:
```
except ImportError as exc:
    raise ImproperlyConfigured('The Redis backend requires redis-py to be installed.') from exc
```

constance/backends/redisd.py
sergei-iurchenko: This merge request is dedicated to linters so let`s leave it.
I agree that pickle is not the best choice.

constance/models.py
```diff
@@ -9,7 +9,7 @@
         "Couldn't find the the 3rd party app "
         'django-picklefield which is required for '
         'the constance database backend.'
-    )
+    ) from None
```
sergei-iurchenko, line 12: the same as above

example/cheeseshop/settings.py
```diff
@@ -79,7 +79,6 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        #'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
```
sergei-iurchenko, line 82: redundant line

sergei-iurchenko (commented):
new breath in old codebase!

tests/backends/test_database.py
Mogost: I find the rule useful.

constance/apps.py
Mogost: It shouldn't. It's not required for checks. Discussed in https://github.com/jazzband/django-constance/pull/562#discussion_r1664732150

constance/backends/redisd.py
Mogost: Here from exc does not provide for user more readable exception.  If import error occurred you actually do not need more than "The Redis backend requires redis-py to be installed"

tests/backends/test_database.py
sergei-iurchenko: why? what is the sense of assigment?

constance/apps.py
sergei-iurchenko: I have checked this case attentively. I have just set breakpoint at the beginning of checks.py file.
If we remove import from apps.py then check no matter works.
But it works because of import in forms.py **from .checks import get_inconsistent_fieldnames**!
Thus correct work of checks is just a side effect of this import. If I move **import from .checks import get_inconsistent_fieldnames** from module import to  local in **def clean(self)** function then checks are not run.
Future refactoring can break checks in implicit manner.
I suggest to return this import to the place where it was.
Also it is recommended way by django documentation https://docs.djangoproject.com/en/5.0/topics/checks/#registering-and-labeling-checks

constance/apps.py
sergei-iurchenko: I resolve this to return to original comment

example/cheeseshop/settings.py
sergei-iurchenko: my fault(

constance/apps.py
Mogost: Thank you. Fixed https://github.com/jazzband/django-constance/pull/562/commits/68a1c5036e87b9ce88f67b8cd6028197ddb85572

tests/backends/test_database.py
Mogost: Refactored https://github.com/jazzband/django-constance/pull/562/commits/3d4652c2a140d2dcf58d2ab01a699e9e56343c60

constance/apps.py
```diff
@@ -1,11 +1,14 @@
...
+
 
 class ConstanceConfig(AppConfig):
     name = 'constance'
     verbose_name = _('Constance')
     default_auto_field = 'django.db.models.AutoField'
 
     def ready(self):
-        from . import checks
+        checks.register(check_fieldsets, 'constance')
```
sergei-iurchenko, line 14: There is no need to register check here. They are registered automatically in decorator while loading module. Only import is required here.

sergei-iurchenko (approved):

constance/apps.py
Mogost: I just reused practice from django. 

https://github.com/django/django/blob/f302343380c77e1eb5dab3b64dd70895a95926ca/django/contrib/sites/apps.py#L17

I'm not too fond of import which is not clearly understandable.

```
import this
The Zen of Python, by Tim Peters
...
Explicit is better than implicit.
```

Mogost:
@sergei-iurchenko @Natim thank you for the review! 🦸‍♂️

constance/checks.py
```diff
@@ -1,14 +1,11 @@
...
-from typing import Set
-from typing import Tuple
+from __future__ import annotations
 
 from django.core import checks
 from django.core.checks import CheckMessage
 from django.utils.translation import gettext_lazy as _
 
 
-@checks.register('constance')
```
Natim, line 10: :pray:

PR #571, opened 2024-08-23
Fix #570

### LINKED ISSUE #570: Missing migration in 4.0.0
### Describe the problem

When using django-constance with django>5, the command to check for missing migrations (`makemigrations --check --dry-run`) tells me that there is a missing migration. I guess this is because of the `editable` field, which is set here: https://github.com/jazzband/django-constance/blob/master/constance/migrations/0001_initial.py#L16, but it is not stated on the model (https://github.com/jazzband/django-constance/blob/master/constance/models.py#L7).

The created migration looks like this:
```
# Generated by Django 5.1 on 2024-08-23 15:01

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('constance', '0003_drop_pickle'),
    ]

    operations = [
        migrations.AlterField(
            model_name='constance',
            name='value',
            field=models.TextField(blank=True, null=True),
        ),
    ]
```    

In case you can reproduce this, I am happy to help.

### Steps to reproduce

* In the example-project, install Django > 5 and run `python manage.py makemigrations --check --dry-run`
* Run tox as configured here: https://github.com/sebastianmanger/django-constance/commit/2357a61e64f698f18189508d0317806092f5c0f0. Example output: https://github.com/sebastianmanger/django-constance/actions/runs/10527827693/job/29171898640#step:5:542

### System configuration

* Django version: >5
* Python version: any
* Django-Constance version: latest (e428387)

Mogost:


camilonova (approved):

Mogost:
@camilonova could you please approve a new release with this fix. https://github.com/jazzband/django-constance/releases/tag/4.0.1
I think this is crucial.

PR #579, opened 2024-09-03
Fix exception in 0002 migration

Mogost:
@sergei-iurchenko could you please check this improvement?

### DIFF
--- a/constance/migrations/0002_migrate_from_old_table.py
+++ b/constance/migrations/0002_migrate_from_old_table.py
@@ -1,7 +1,6 @@
 from logging import getLogger
 
 from django.core.management.color import no_style
-from django.db import DatabaseError
 from django.db import migrations
 
 logger = getLogger(__name__)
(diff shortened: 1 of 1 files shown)


constance/migrations/0002_migrate_from_old_table.py
```diff
@@ -14,15 +13,16 @@ def _migrate_from_old_table(apps, schema_editor) -> None:
...
     quoted_string = ', '.join([connection.ops.quote_name(item) for item in ['id', 'key', 'value']])
-    try:
-        with connection.cursor() as cursor:
-            cursor.execute(
-                f'INSERT INTO constance_constance ( {quoted_string} ) SELECT {quoted_string} FROM constance_config',  # noqa: S608
-                [],
-            )
-            cursor.execute('DROP TABLE constance_config', [])
-    except DatabaseError:
-        logger.exception('copy data from old constance table to a new one')
```
sergei-iurchenko, line 25: i think you can just change logger.exception -> logger.info and that would make error not verbose

ivan-klass: I believe idea here is not to suppress the logger but don't hide any other errors except already missing table - i.e. all the connection errors, permission errors, etc. 
In theory, listing table names may require a specific permission which can be configured separate from SELECT / INSERT / UPDATE / ALTER on some tables, but that seems to be an uncommon and impractical thing

Mogost: Yep, exactly. I think suppressing an exception is dangerous and can lead to many problems which we can avoid right here.

sergei-iurchenko (approved):

PR #581, opened 2024-09-03
Collections support

### LINKED ISSUE #574: Add support for collection types (list/tuple/dict/set/etc.)
### Describe the problem

Using the `DatabaseBackend`, I store lists of recipients for emails and dicts for small configy-things. The (awesome) changes in https://github.com/jazzband/django-constance/pull/564 do not cover these types. The stored value in the database is not in the expected format
```
'{"__type__": "default", "__value__": "{"test": "test"}"}'
```
but 
```
'{"test": "test"}'
```

This then leads to an exception when trying to access the value, as the keys do not match the expected structure (https://github.com/jazzband/django-constance/blob/master/constance/codecs.py#L58).

**Proposed discussion**

I could help out with any of the following tasks:

* [ ] treat the defined types (https://github.com/jazzband/django-constance/blob/master/constance/codecs.py#L37 and https://github.com/jazzband/django-constance/blob/master/constance/codecs.py#L83) as "whitelist" and raise on serialization of unknown types.
* [ ] document the supported types, or indicate the internal wokings. A simple sentence indicating usage of JSON as data structure would probably be good enough.
* [ ] try to deserialize the object (https://github.com/jazzband/django-constance/blob/master/constance/codecs.py#L57) with the default JSONSerializer before raising an exception. This might be helpful if other users face the same issue and have already-migrated data.
* [ ] probably out of scope: use the registered types for nested serialization (e.g. a dict with a UUID).
 
### Steps to reproduce

This simply uses the default encoding defined in the `JSONEncoder`, but should display the issue clearly 😄 
See https://github.com/jazzband/django-constance/compare/master...sebastianmanger:django-constance:json-missing-types

### System configuration

* Django version: any
* Python version: any
* Django-Constance version: >4.0.0

Mogost:
Fix #574

@sebastianmanger @lpomfrey This PR should address your problem from #574, could you please review/test it?

cc: @ivan-klass @Natim @camilonova

### DIFF
--- a/constance/codecs.py
+++ b/constance/codecs.py
@@ -34,14 +34,20 @@ def _as(discriminator: str, v: Any) -> dict[str, Any]:
 def dumps(obj, _dumps=json.dumps, cls=JSONEncoder, default_kwargs=None, **kwargs):
     """Serialize object to json string."""
     default_kwargs = default_kwargs or {}
-    is_default_type = isinstance(obj, (str, int, bool, float, type(None)))
+    is_default_type = isinstance(obj, (list, dict, str, int, bool, float, type(None)))
     return _dumps(
         _as(DEFAULT_DISCRIMINATOR, obj) if is_default_type else obj, cls=cls, **dict(default_kwargs, **kwargs)
     )
 
 
-def loads(s, _loads=json.loads, **kwargs):
+def loads(s, _loads=json.loads, *, first_level=True, **kwargs):
     """Deserialize json string to object."""
+    if first_level:
+        return _loads(s, object_hook=object_hook, **kwargs)
+    if isinstance(s, dict) and '__type__' not in s and '__value__' not in s:
+        return {k: loads(v, first_level=False) for k, v in s.items()}
+    if isinstance(s, list):
+        return list(loads(v, first_level=False) for v in s)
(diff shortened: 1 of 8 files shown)


lpomfrey:
We've done some testing and this would seem to fix our problems from #574 :+1:

camilonova (approved):

camilonova:
Looks good to me

Mogost:
There's a lack of documentation. If anyone has time to help with it, I would be glad!

sebastianmanger:
Thanks a lot! This fully supports "my" use cases.
As a "pure" user of the library, I personally would expect something minimal like https://github.com/sebastianmanger/django-constance/commit/7c24c6de576e29eaf8aa9b2f449331bc1ecea3e0

Mogost:
> Thanks a lot! This fully supports my use cases. As a pure user of the library, I personally would expect something minimal like [sebastianmanger@7c24c6d](https://github.com/sebastianmanger/django-constance/commit/7c24c6de576e29eaf8aa9b2f449331bc1ecea3e0)

Thank you. cherry-picked

Mogost:
Tagged with 4.1.0 version. Should be released soon

PR #604, opened 2025-01-07
fix(docs): use sphinx note function

Rotzbua:
Instead of text "note" use the function which sphinx offers.


camilonova (approved):

camilonova:
Thank you

PR #607, opened 2025-01-11
Add get_values_for_keys function to utils

### LINKED ISSUE #344: [Improvement] Option for multiget values
### Describe the improvement

I think it would be a nice feature to have the option to use the [mget](https://github.com/jazzband/django-constance/blob/master/constance/backends/redisd.py#L41) method that both backends already implements inside the config.
We have several cases where we call a few keys in the same operation. If we could make a single request (in our cause to redis), it would decrease our hits to redis by a lot.

Think it's a valid idea? The hardest part (the implementation of mget itself on both backends) is already done. Just would need to let the method visible inside the Config.

FernandoKGA:
Closes #344.

### DIFF
--- a/constance/utils.py
+++ b/constance/utils.py
@@ -20,3 +20,27 @@ def get_values():
     default_initial = ((name, options[0]) for name, options in settings.CONFIG.items())
     # Then update the mapping with actually values from the backend
     return dict(default_initial, **dict(config._backend.mget(settings.CONFIG)))
+
+def get_values_for_keys(keys):
+    """
+    Retrieve values for specified keys from the backend.
+
+    :param keys: List of keys to retrieve.
+    :return: Dictionary with values for the specified keys.
+    :raises AttributeError: If any key is not found in the configuration.
+    """
+    if not isinstance(keys, (list, tuple, set)):
+        raise TypeError('keys must be a list, tuple, or set of strings')
+
+    # Prepare default initial mapping
+    default_initial = {
+        name: options[0] for name, options in settings.CONFIG.items() if name in keys
+    }
+
+    # Check if all keys are present in the default_initial mapping
+    missing_keys = [key for key in keys if key not in default_initial]
+    if missing_keys:
+        raise AttributeError(f'Key "{missing_keys[0]}" not found in configuration.')
+
+    # Merge default values and backend values, prioritizing backend values
+    return dict(default_initial, **dict(config._backend.mget(keys)))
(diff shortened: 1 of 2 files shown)


FernandoKGA:
@camilonova and @Mogost , could you guys take a look? In my opinion it solves the issue, not inside the Config as mentioned, but definitely a use case that I would particularly appreciate to reduce database access.

constance/utils.py
```diff
@@ -20,3 +20,27 @@ def get_values():
...
+
+    # Prepare default initial mapping
+    default_initial = {
+        name: options[0] for name, options in settings.CONFIG.items() if name in keys
+    }
+
+    # Check if all keys are present in the default_initial mapping
+    missing_keys = [key for key in keys if key not in default_initial]
+    if missing_keys:
+        raise AttributeError(f'Key "{missing_keys[0]}" not found in configuration.')
```
Natim, line 43:
```suggestion
        raise AttributeError(f'"{", ".join(missing_keys)}" keys not found in configuration.')
```

constance/utils.py
```diff
@@ -20,3 +20,27 @@ def get_values():
     default_initial = ((name, options[0]) for name, options in settings.CONFIG.items())
     # Then update the mapping with actually values from the backend
     return dict(default_initial, **dict(config._backend.mget(settings.CONFIG)))
+
+def get_values_for_keys(keys):
```
Natim, line 24:
```suggestion
def get_values_for_keys(keys: list):
```

Natim:
The PR looks good to me, both comments are just nits

tests/test_utils.py
```diff
@@ -62,3 +62,24 @@ def test_get_values(self):
...
+    def test_get_values_for_keys_empty_keys(self):
+        result = get_values_for_keys([])
+        self.assertEqual(result, {})
+
+    def test_get_values_for_keys_throw_error_if_no_key(self):
+        self.assertRaisesMessage(AttributeError, 'Key "OLD_VALUE" not found in configuration.', get_values_for_keys, ['BOOL_VALUE', 'OLD_VALUE'])
+
+    def test_get_values_for_keys_invalid_input_type(self):
+        with self.assertRaises(TypeError):
+            get_values_for_keys("key1")
```
Natim, line 85: missing new line at the end of the file.

constance/utils.py
FernandoKGA: This you restrict the type to `list`. Should I check only for lists?

While writing, I've thought one case that maybe worth testing: pass the same value multiple times to the list and see what happens. What do you think?

Maybe expect `set` only is a better restriction?

FernandoKGA: Enforcing a `set` would avoid the library to check duplicated keys as it can happen in a `list`.

Mogost: Iterable[str] looks like the right typing here

Mogost (approved):

PR #609, opened 2025-01-27
Fix override_config test decorator on Django 5.2

pfouque:
Yet another Django 5.2 fix:

Django 5.2 is the first version to ship [this PR](https://github.com/django/django/pull/18514/files) turning `pre_setup` into a **classmethod** and generating this error (already visible in tox, but ignored)

```
  ERROR: test_override_config_on_class_changes_config_value (tests.test_test_overrides.OverrideConfigClassDecoratorTestCase.test_override_config_on_class_changes_config_value)
  Assert that the class decorator changes config.BOOL_VALUE.
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/home/runner/work/django-constance/django-constance/constance/test/unittest.py", line 49, in _pre_setup
      original_pre_setup(inner_self)
  TypeError: TransactionTestCase._pre_setup() takes 1 positional argument but 2 were given
```

I found this fix, following the current solution but maybe a better (and more modern) solution exists
"""
Override the config by modifying TestCase methods.

This method follows the Django <= 1.6 method of overriding the
_pre_setup and _post_teardown hooks rather than modifying the TestCase
itself.
"""

### DIFF
--- a/constance/test/unittest.py
+++ b/constance/test/unittest.py
@@ -46,10 +46,12 @@ class override_config(override_settings):
         original_post_teardown = test_case._post_teardown
 
         if DJANGO_VERSION < (5, 2):
+
             def _pre_setup(inner_self):
                 self.enable()
                 original_pre_setup(inner_self)
         else:
+
             @classmethod
             def _pre_setup(cls):
                 # NOTE: Django 5.2 turned this as a classmethod


constance/test/unittest.py
```diff
@@ -44,9 +45,19 @@ def modify_test_case(self, test_case):
...
-            self.enable()
-            original_pre_setup(inner_self)
+        if DJANGO_VERSION < (5, 2):
+
+            def _pre_setup(inner_self):
+                self.enable()
+                original_pre_setup(inner_self)
+        else:
+
+            @classmethod
```
Natim, line 55: Can we create a decorator for that so that we don't have to duplicate the code of the method?

pfouque: The decorator is not the only difference, the call to original_pre_setup also differs (as we don't have to pass the instance anymore)
Additionally, this version is probably easier to clean up when Django5.2 will be the minimum supported version

Mogost (approved):

pfouque:
Thanks @Natim & @Mogost !

PR #613, opened 2025-03-05
feat: Add generated field - derived_value

wtlyu:
## Overview
Derived value fields are fields that are calculated based on other fields. They are read-only and not visible/editable in the admin interface.

## Defining Derived Value Fields
To define a derived value field, use the `derived_value` type in the `CONSTANCE_CONFIG` tuple, and provide a callable that returns the desired value as the default. The callable will be invoked with the config object as the only argument, allowing you to access other config values (e.g., `constance.config`). 

Be cautious when using the `config` object to access the current value of the derived field to avoid creating infinite loops.

## Valid Callables for Derived Value Fields
The following are valid default parameters for derived value fields:
- A function object
- A string containing the path to a function object
- A lambda function

### Example

```python
def get_answer(config):
    return 'The answer is %s' % config.THE_ANSWER

CONSTANCE_CONFIG = {
    'THE_ANSWER': (42, 'Answer to the Ultimate Question of Life, '
                        'The Universe, and Everything'),
    'THE_QUESTION': ('What is the answer to the ultimate question of life, the universe, and everything?', 'The question'),
    'THE_QUESTION_ANSWERED_LAMBDA': (lambda config: 'The answer is %s' % config.THE_ANSWER, 'The question answered', 'derived_value'),
    'THE_QUESTION_ANSWERED_FUNCTION': (get_answer, 'The question answered', 'derived_value'),
    'THE_QUESTION_ANSWERED_FUNCTION_PATH': ('path.to.get_answer', 'The question answered', 'derived_value'),
}
```

ResultsThis will yield the following values:
- THE_QUESTION_ANSWERED_LAMBDA: `The answer is 42`
- THE_QUESTION_ANSWERED_FUNCTION: `The answer is 42`
- THE_QUESTION_ANSWERED_FUNCTION_PATH: `The answer is 42`

**Important Note**: Derived value fields are not editable in the admin interface, and their value is recalculated every time the config object is accessed. Consequently, derived value fields should never be included in `CONSTANCE_CONFIG_FIELDSETS`.

### DIFF
--- a/constance/admin.py
+++ b/constance/admin.py
@@ -104,6 +104,8 @@ class ConstanceAdmin(admin.ModelAdmin):
             django_version=get_version(),
         )
         for name, options in settings.CONFIG.items():
+            if len(options) == 3 and options[2] == 'derived_value':
+                continue
             context['config_values'].append(self.get_config_value(name, options, form, initial))
 
         if settings.CONFIG_FIELDSETS:
--- a/constance/base.py
+++ b/constance/base.py
@@ -1,6 +1,12 @@
 from . import settings
 from . import utils
+import importlib
 
+
+def get_function_from_string(path):
+    module_path, function_name = path.rsplit('.', 1)
+    module = importlib.import_module(module_path)
+    return getattr(module, function_name)
 
 class Config:
     """The global config wrapper that handles the backend."""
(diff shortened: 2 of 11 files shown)


sergei-iurchenko:
Why do you want derived values in constance instead of using function in your code?

Issue #633, opened 2025-09-19
Management command to remove stale keys doesn't respect CONSTANCE_DATABASE_PREFIX


christherama:
## Description

When running the `constance` management command with `remove_stale_keys`, the Django setting `CONSTANCE_DATABASE_PREFIX` is not honored. That means that using the [DatabaseBackend](https://github.com/jazzband/django-constance/blob/6970708e05d769b2e424c4c22790eb2fd778ca88/constance/backends/database.py#L18) with a non-empty string prefix will result in all keys being dropped from the database. If `CONSTANCE_DATABASE_CACHE_BACKEND` is set to `False`, or if caching is enabled but the TTL of any item has expired, then the next usage of any configuration item will have its default value.

## Steps to Reproduce

1. In a Django project with `django-constance>2.6`, use a non-empty string prefix:
   ```python
   # your/project/settings.py, where DJANGO_SETTINGS_MODULE=your.project.settings
   
   CONSTANCE_BACKEND = "constance.backends.database.DatabaseBackend"
   CONSTANCE_DATABASE_CACHE_BACKEND = None
   CONSTANCE_DATABASE_PREFIX = "constance:"
   CONSTANCE_CONFIG = {
     "MY_FEATURE_ENABLED": (
       True,
       "Whether my feature is enabled",
     )
   }
   ```
2. In a terminal, start your database and cache
3. Start the dev server with `python manage.py runserver`
4. Navigate to the Django Admin Constance page and update `MY_FEATURE_ENABLED` to `False`, being sure to save
5. In a terminal, run the management command:
   ```shell
   $ python manage.py constance remove_stale_keys
   The following record will be deleted:
   constance:MY_FEATURE_ENABLED	{"__type__": "default", "__value__": false}
   ```
6. Refresh the Django Admin Constance page. You'll see the default value of `True` instead of the value previously set in step 4

PR #634, opened 2025-09-20
Respects database prefix when removing stale keys

christherama:
This PR updates the management command to respect the configured `CONSTANCE_DATABASE_PREFIX` when removing stale keys. This is a fix for https://github.com/jazzband/django-constance/issues/633

### DIFF
--- a/constance/management/commands/constance.py
+++ b/constance/management/commands/constance.py
@@ -70,7 +70,8 @@ class Command(BaseCommand):
             for k, v in get_values().items():
                 self.stdout.write(f'{k}\t{v}', ending='\n')
         elif command == self.REMOVE_STALE_KEYS:
-            actual_keys = settings.CONSTANCE_CONFIG.keys()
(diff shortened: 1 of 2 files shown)


constance/management/commands/constance.py
```diff
@@ -70,7 +70,8 @@ def handle(self, command, key=None, value=None, *args, **options):
             for k, v in get_values().items():
                 self.stdout.write(f'{k}\t{v}', ending='\n')
         elif command == self.REMOVE_STALE_KEYS:
-            actual_keys = settings.CONSTANCE_CONFIG.keys()
+            prefix = getattr(settings, 'CONSTANCE_DATABASE_PREFIX', '')
+            actual_keys = [f'{prefix}{key}' for key in settings.CONSTANCE_CONFIG]
```
Natim, line 74: Should we use the backend `add_prefix` utility for that?

PR #641, opened 2025-10-07
Run `ruff format`

Natim:


pyproject.toml
```diff
@@ -64,7 +64,6 @@ line-length = 120
 indent-width = 4
 
 [tool.ruff.format]
-quote-style = "single"
```
Mogost, line 67: https://github.com/jazzband/django-constance/pull/560#discussion_r1663932544

Do we really need to change style?

Natim: It feels odd to enforce single quotes while the Python guidelines is using double quotes everywhere else, doesn't it?

Natim: Thank you for reminding me of the discussion, I believe one year later we can do the change since we both agree on this.

PR #668, opened 2026-05-31
Remove idempotent autofill to fix lazy config load in async context

### LINKED ISSUE #667: `django-constance` lazy init triggers `SynchronousOnlyOperation` when first accessed in async context
The `constance.config` initializes itself on first attribute access. Initialization performs a synchronous cache read against the database. If the first access to `config` happens inside an async context (e.g. an ASGI request handler), Django raises:

```
django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
```

Workaround: Pre-initialize constance with `config._setup()`.

Reproduction script:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "django>=4.2,<5.0",
#     "django-constance==4.3.5",
# ]
# ///
from __future__ import annotations

import asyncio
import os
import sys
import tempfile

import django
from constance import config
from django.conf import settings

DB_PATH = os.path.join(tempfile.mkdtemp(prefix="constance-repro-"), "db.sqlite3")

settings.configure(
    DATABASES={
        "default": {
            "ENGINE": "django.db.backends.sqlite3",
            "NAME": DB_PATH,
        },
    },
    INSTALLED_APPS=[
        "constance",
    ],
    CACHES={
        "default": {
            "BACKEND": "django.core.cache.backends.db.DatabaseCache",
        },
    },
    CONSTANCE_BACKEND="constance.backends.database.DatabaseBackend",
    CONSTANCE_DATABASE_CACHE_BACKEND="default",
    CONSTANCE_CONFIG={
        "EXAMPLE_FLAG": (True, "An example feature flag", bool),
    },
)

django.setup()

from django.core.management import call_command  # noqa: E402

call_command("migrate", verbosity=0)
call_command("createcachetable", verbosity=0)


def main() -> int:
    if "--workaround" in sys.argv:
        config._setup()  # noqa: SLF001

    async def read_flag() -> bool:
        return bool(await config.EXAMPLE_FLAG)

    asyncio.run(read_flag())


if __name__ == "__main__":
    sys.exit(main())
```

```bash
# Reproduces the bug (raises SynchronousOnlyOperation):
uv run script.py

# Confirms the workaround (exits cleanly):
uv run script.py --workaround
```

PhilippTh:
Fixes #667: lazy access to `constance.config` from an async context raised `SynchronousOnlyOperation` because `DatabaseBackend.__init__` called `self.autofill()`, triggering a sync cache/DB read before `Config.__getattr__`'s async-aware branch could run.

The autofill call is removed from `__init__` — it's already invoked lazily on cache miss in `get()` and `aget()`, so the warm-up happens on first read instead of at backend construction. Same total I/O, just deferred.

## Tests

- New regression tests in `TestDatabaseBackendConstruction` assert `DatabaseBackend()` performs zero queries, with and without a cache backend configured. Verified to fail on master without the fix.
- New end-to-end test `test_lazy_config_first_access_in_async` exercises `await LazyConfig().INT_VALUE` from inside the event loop.

### DIFF
--- a/constance/backends/database.py
+++ b/constance/backends/database.py
@@ -40,7 +40,6 @@ class DatabaseBackend(Backend):
                 )
         else:
             self._cache = None
-        self.autofill()
         # Clear simple cache.
         post_save.connect(self.clear, sender=self._model)
 
--- a/tests/backends/test_database.py
+++ b/tests/backends/test_database.py
@@ -72,6 +72,11 @@ class TestDatabaseBackendConstruction(TestCase):
         with self.assertNumQueries(0):
             DatabaseBackend()
 
+    def test_autofill_is_noop_without_cache(self):
+        backend = DatabaseBackend()
+        with self.assertNumQueries(0):
+            backend.autofill()
+
     def test_backend_init_does_no_queries_with_cache(self):
         old_cache_backend = settings.DATABASE_CACHE_BACKEND
         settings.DATABASE_CACHE_BACKEND = "default"
--- a/tests/backends/test_database.py
+++ b/tests/backends/test_database.py
@@ -1,7 +1,9 @@
 from django.test import TestCase
 from django.test import TransactionTestCase
 
+from constance import LazyConfig
 from constance import settings
+from constance.backends.database import DatabaseBackend
 from constance.base import Config
 from tests.storage import StorageTestsMixin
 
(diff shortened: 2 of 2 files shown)


PhilippTh:
@camilonova or @Mogost, would you be able to have a look at this as well? Thanks!

PR #669, opened 2026-06-15
Made 0003_drop_pickle.py database (multi-tenant) aware

janssen70:
migrate_pickled_data() always operates on default database, ignoring the --database <dbname> argument of the migrate command in multi-tenant setups. Migration is then marked as complete in database <dbname>, causing an inconsistency that prevents the tenant from accessing constance data.

Found after upgrading from 3.1.0 to 4.3.5, using Django 6.0.6
I think provided change should avoid this problem. To be open with you, I didn't run tests. 
Hope this helps


constance/migrations/0003_drop_pickle.py
```diff
@@ -29,11 +29,12 @@ def import_module_attr(path):
 
 def migrate_pickled_data(apps, schema_editor) -> None:  # pragma: no cover
     Constance = apps.get_model("constance", "Constance")
+    db_alias = schema_editor.connection.alias
```
camilonova, line 32: What's the default value when it's not multi-tenant?

janssen70: Then it refers to the default database, see https://docs.djangoproject.com/en/6.0/ref/schema-editor/#connection

janssen70: My original comment is not as clear as I thought. let me retry. The problem is that when migrating a nondefault database, so using e.g. --database tenant1_db, the data in the default database get migrated instead. Then, the migration is marked as complete in the nondefault database, here tenant1_db, as it should. Migrations seem o.k. but accessing constance parameters doesn't work.

PR #672, opened 2026-08-10
Prevent read-only form state from leaking between requests

gpg4:
## Problem

ConstanceForm reuses field constructor option dictionaries from the module-level FIELDS mapping. When a request user does not have constance.change_config, the form adds disabled=True directly to one of those shared dictionaries.

In a long-running process, that state survives after the read-only form has been built. Later forms of the same field type remain disabled even when their request user has change permission.

## Fix

Copy the field options before applying request-specific read-only state. This keeps disabled scoped to the current form construction and leaves the shared FIELDS definitions unchanged.

A regression test builds a read-only form followed by an editable form and verifies that only the first form is disabled.

## Testing

- django-admin test: 154 passed, 1 skipped
- focused pytest tests/test_form.py: 2 passed
- django-admin makemigrations --check --dry-run: no changes detected
- Ruff 0.5.0 lint: passed
- Ruff 0.5.0 format check: passed


camilonova (approved):
LGTM

